PR #2, opened 2021-04-07
Fix tatuylonen/wikitextprocessor#1

### LINKED ISSUE #1: Error using transient templates
When using transient templates, I get a Python error about the "body" variable being used before it's declared. I believe in the line linked below, the index should reference "name" and not "body".

https://github.com/tatuylonen/wikitextprocessor/blob/b092f0038bb39670ecc2687f2a4d6bcc80361f40/wikitextprocessor/core.py#L1081

fpw:
This should fix #1


tatuylonen (approved):

tatuylonen:
Thanks!  I just merged the fix.

Issue #4, opened 2021-12-07
wikitextprocessor is not suitable to parse non-english Wikipedia


GaaH:
Currently, wikitextprocessor detects template by checking if a page’s title starts with "Templates:". This strategy works with english Wikipedia, however it doesn’t with other wikis. Indeed, frwiki’s template pages start with "Modèle:", hence templates from frwiki are not properly detected.

A solution (maybe THE solution) would be to rely on the _ns_ tag in pages. For instances, pages with ns = 10 are templates pages. Another solution would be to get the prefix of namespace 10, which can be found in the _namespaces_ tag at the beginnig of an xml dump.

tatuylonen:
Your proposed solution seems reasonable.  I'll try to implement it this week (as soon as I get inflection table parsing in wiktextract out of the way).

GaaH:
I’m currently trying to implement something, if I achieve to make it works I’ll submit a PR.
But we need to make sure that every wiki use the key 10 to refer to namespaces. I think it is the case, but I’m not sure…

tatuylonen:
I'll be happy to merge if you get it implemented.

GaaH:
I did some edit to your code, you can find it on [this branch](https://github.com/GaaH/wikitextprocessor/tree/namespaces). I mostly changed how parsing a mediawiki dump is done. For instance, I switch to xml.etree to parse the xml, because the regex used to parse xml tag’s attr was not working as intended. I don’t know if it hurts speed though, but I’m sure parsing will be a lot more reliable. If you want, I can submit a PR with only the switch from regex parsing to xml.etree, without touching too much the rest of the code. If not, is there any reason not to rely on xml parsers ?

I also added some classes to represent a page, since namespace information should be part of the page attributes.

All tests are green, so I guess I did not break too much things, but I do not understand all the processes taking place in the file core.py, so I’m having troubles editing this file for now.

tatuylonen:
I used to use one of the existing xml parsers but switched to regexps at some point.  I remember having some issues that triggered the change, but I can't now remember what exactly the issue was.  I was optimizing performance at the same time and I think the regexp approach was moderately faster.  At least would be good to compare if there is noticeable difference - though I think that there is now so much more other processing that the XML parsing speed isn't very significant.  "ns" field was listed as one of the ignored fields so if still using the regexp-based parser it would need to be removed from the ignored list.

GaaH:
[...]

PR #9, opened 2022-08-04
Polyfill for removed Lua functions

garfieldnate:
While attempting to parse out the Simple English wiktionary (related to #4), I hit errors in Lua when executing templates. The issue was that Wikipedia uses Lua 5.1 everywhere, and Lua 5.3 removed some functions from `table`. In this PR I only polyfill `table.maxn`, though it's possible we will encounter a few others in the future. This allows us to parse the Simple English wiktionary without errors.

### DIFF
--- a/wikitextprocessor/core.py
+++ b/wikitextprocessor/core.py
@@ -63,15 +63,16 @@ def phase2_page_handler(dt):
             assert isinstance(data, str)
         else:
             data = None
-        try:
-            ret = _global_page_handler(model, title, data)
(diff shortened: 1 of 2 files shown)


garfieldnate:
I've removed https://github.com/tatuylonen/wikitextprocessor/pull/9/commits/27cf2554059eea7bc46d8fa8a5e34b831726a5cc, although this commit was very important in debugging the issue. I'm not sure what causes `traceback.format_exception` to give incomplete stack traces.

PR #13, opened 2022-09-15
Move namespace text to JSON files

### LINKED ISSUE #4: wikitextprocessor is not suitable to parse non-english Wikipedia

xxyzz:
This pull request resolves #4. It gets the language and namespace data from MediaWiki API, then save them to JSON files for later use. The namespace API works for all languages. But the language API is only available for English and Chinese Wiktionary, other language JSON files would need to be created manually.

I can also add a GitHub Actions job to run the tests in another pull request if you want that.


tatuylonen:
Just a heads-up, I'm looking into this.  I want to do a bit more review & testing before merging, but it looks like you've made significant progress and I look forward to merging your changes.  Great news that you've gotten Chinese Wiktionary to parse, even if it is not completely processed yet.

tatuylonen:
I also want to say that I really appreciate your contributions.  I did a quick review of the changes in this pull request, and generally it looks very good to me.

There is going to be a lot of language-specific configuration on the wiktextract side.  I'm thinking maybe it would be better to put all the language-specific / Wiktionary edition specific configuration in one place (on the wiktextract side) and pass the necessary information down to wikitextprocessor.  What do you think?  (I don't think this prevents a merge - we can easily change where the information is configured later.)

xxyzz:
I could move the "get_languages.py" and those "lanuages.json" files to wiktextract, load the JSON file in wiktextract then pass the dict object to `Wtp`. Only the "luaxec.py" file uses this data, probably only need to change a few lines.

tatuylonen:
These (and the corresponding changes in wiktextract) have now been merged.  Thank you for the great contributions!!!

PR #16, opened 2022-11-14
Remove LanguageConverter markups after all templates are expanded

xxyzz:
The previous code would break normal templates, for example: `This ending is semi-{{glossary|productive}}` in https://en.wiktionary.org/wiki/-ant#German.

### DIFF
--- a/wikitextprocessor/core.py
+++ b/wikitextprocessor/core.py
@@ -1353,6 +1353,11 @@ class Wtp:
 
         # Expand any remaining magic cookies and remove nowiki char
         expanded = self._finalize_expand(expanded)
+
+        # Remove LanguageConverter markups: https://www.mediawiki.org/wiki/Writing_systems/Syntax
+        if not pre_expand and "-{" in expanded:
+            expanded = expanded.replace("-{", "").replace("}-", "")
+
         return expanded
 
     def _finalize_expand(self, text):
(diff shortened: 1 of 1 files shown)


kristian-clausal:
The previous code would not break normal templates, unless the template is in the middle of two -'s: `-{{template}}-`. -ant/German is not affected, for example.

xxyzz:
I mean my previous commit https://github.com/tatuylonen/wikitextprocessor/commit/2df8bff0b0d7f9cb82701f24cd9a516b06e0a3f2 would break templates because "word-{{template}}" will be replaced in the pre expand process but all template should be expanded when `pre_expand` is `False`. Your commit https://github.com/tatuylonen/wikitextprocessor/commit/3c4f5d5ff91feb1aa1a71bf73f8d2879f7dffdd5 actually doesn't use regex...

Regex is not needed here, "-{Chinese word}-" won't be expanded as template, a simple `str.replace()` is enough.

kristian-clausal:
Yeah, I put a regex pattern where it wouldn't work. It only fixed the issues because it stopped the new added removal.

Removing all -{ and }- could mess up some *text* however, which doesn't necessarily have to be a template. It is better to use a complete regex in this case.

xxyzz:
One three pages use "-{" in normal text, all of them are used to represent chemical name:
- https://en.wiktionary.org/wiki/taurocholic_acid
- https://en.wiktionary.org/wiki/sisomicin
- https://en.wiktionary.org/wiki/gentamicin

xxyzz:
And sadly your new commit https://github.com/tatuylonen/wikitextprocessor/commit/c10f9c3ab906f79d307d95ca550cb23b4038f51b still doesn't work... `-{([^}\n])}-` only match one character between "-{" and "}-" and it would remove the hole string like this "-{我}-".

kristian-clausal:
Thank you for pointing out my mistake. I should have copy-pasted the original... I will amend the regex to take into account that -{ and }- should only occur at word boundaries, not within words.

PR #17, opened 2022-11-15
Only remove LanguageConverter marks for Chinese Wiktionary

xxyzz:
There are three pages in English Wiktionary use "-{" and "-}" in normal text to
represent chemical names(Preferred IUPAC name): taurocholic acid, sisomicin, gentamicin

This commit also fix the error in https://github.com/tatuylonen/wikitextprocessor/commit/c10f9c3ab906f79d307d95ca550cb23b4038f51b as mentioned here: https://github.com/tatuylonen/wikitextprocessor/pull/16#issuecomment-1314553543


kristian-clausal:
I've amended the broken regex so that it shouldn't affect chemical formulas (-{ should not appear within words, after all), and I also added checking for langcode.

Because there are three tests to check if the string is even eligible to have substitutions made (langcode, not pre-expanded AND that there's `-{` present) the regex should almost never trigger so there should be no performance issue with string-method vs. regex use.

I didn't notice this pull request so the changes have already been pushed, but the code and results should be equivalent; closing for now, until someone finds another typo in my regex.

xxyzz:
I think you misunderstand the usage of "-{}-" syntax, it's for preventing auto conversion of Simplified and Traditional Chinese characters, so the characters inside the brackets won't be converted and this syntax could be used between words. The goal of this line of code is to only remove the "-{" and "}-" but your regex also remove the text in between.

And as I said before, regex really is not needed here and make this more complicated... This syntax is only used in Chinese Wiktionary so a language code check will be enough to preserve those three chemical names in English Wiktionary. Chinese Wiktionary clearly won't use "-{" in normal text as it's a syntax.

kristian-clausal:
Thank you for the clarification. Because the original issue was with the removal of -{ breaking templates, I had assumed that it was about {{-style syntaxes. And of course I forgot the god damn \1 from the substitution regex. I hate programming.

But also please write longer and more explanatory commit messages in the future.

With the self.lang_code=="zh" check removing -{ your changes should be all right. I will revert

xxyzz:
> But also please write longer and more explanatory commit messages in the future.
> 

I should also add some tests... Do you mind I create a new pull request to add GitHub Actions for both wikitextprocessor and wikextract projects? So each git push and pull request will trigger a GitHub Actions job to run tests.

kristian-clausal:
> I should also add some tests... Do you mind I create a new pull request to add GitHub Actions for both wikitextprocessor and wikextract projects? So each git push and pull request will trigger a GitHub Actions job to run tests.

I'll have to ask Tatu about GitHub Actions (I don't even know if we use them for anything else). In any case, if you add your tests into the test folders, they will be run when kaikki.org is regenerated, which happens about once a day. The tests are run first, so they are done even if the regen itself fails, and I usually (not today, apparently...) also run make test before commits. Usually...

PR #19, opened 2022-11-15
SIMPLY added namespaces.json for all languages

Wyrun:
I added <language_code>/namespaces.json files which I parsed from two files on github. They could be not perfect, but they should do.


kristian-clausal:
Thank you for these! I will merge this, even if there might be some imperfections, just to have some kind of baseline for others to start from if they want to work on other language wiktionaries.

I've also added a wiktextract/tools folder, if you have a script you did this with; you could add it there for when these need to be updated in the future.

xxyzz:
https://github.com/tatuylonen/wikitextprocessor/blob/main/get_namespaces.py can already add namspace data from any MediaWiki site by calling MediaWiki API and format the json file nicely with `indent` and `ensure_ascii` option.

PR #22, opened 2022-11-21
Replaced \u escape sequence in namespaces

Wyrun:


xxyzz:
Could you use the [get_namespaces.py](https://github.com/tatuylonen/wikitextprocessor/blob/main/get_namespaces.py) script to create these files?

Wyrun:
Ah ok, sorry, didn't notice.

PR #32, opened 2023-03-24
Expand French Wiktionary language and POS heading templates

empiriker:
Similarly to @xxyzz 's approach in #15, parsing the French Wiktionary benefits from dealing with language and pos templates directly before calling `fix_subtitle_hierarchy()` at [here](https://github.com/tatuylonen/wiktextract/blob/093b6892431f1a8e796b18d17fbd8197e74ad0ba/wiktextract/page.py#L2830).

This pull request will support parsing the French Wiktionary via wiktextract (see [pull request](https://github.com/tatuylonen/wiktextract/pull/223) there).

### DIFF
--- a/wikitextprocessor/core.py
+++ b/wikitextprocessor/core.py
@@ -1321,8 +1321,6 @@ class Wtp:
 
                     if self.lang_code == "zh":
                         t = overwrite_zh_template(name, t)
-                    if self.lang_code == "fr":
-                        t = overwrite_fr_template(name, t)
 
                     assert isinstance(t, str)
                     self.expand_stack.pop()  # template name
@@ -1687,33 +1685,5 @@ def overwrite_zh_template(template_name: str, expanded_template: str) -> str:
     elif template_name == "CC-CEDICT":
         # Avoid pasring this license template
         expanded_template = ""
-
-    return expanded_template
-
-def overwrite_fr_template(template_name: str, expanded_template: str) -> str:
-    """
-    Modify some French Wiktionary templates to standard heading format
-    """
-
-    if template_name == "langue":
-        if "sectionlangue" in expanded_template:
-            lang_heading = re.search(r">([^<]+)\[\[", expanded_template).group(1)
-            if "Portail" in lang_heading:
-                lang_heading = re.search(
-                    r"Portail:([^\|]+)\|", expanded_template
-                ).group(1)
-            expanded_template = f"=={lang_heading}=="
-
-    if template_name == "S":
-        regex = re.search(r">([^<]+)<\/span>", expanded_template)
-        if regex is None:
-            return expanded_template
-
-        section_heading = regex.group(1)
-
-        if "titredef" in expanded_template:
-            expanded_template = f"===={section_heading}===="
-        else:
-            expanded_template = f"==={section_heading}==="
 
     return expanded_template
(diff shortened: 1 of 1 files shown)


kristian-clausal:
I'll take a look at these next week, and merge when you feel the code is functional.

One thing which should be thought about (which I don't want to do because it's a headache) is how to decouple wikitextprocessor from wiktionary-parsing stuff. Having language-specific code in isn't agnostic, but the problem is that different wiki-projects have different wikitext things going on...

empiriker:
As I said, I primarily followed xxyzz's approach which is why I made this change to wikitextprocessor. 

However, I ended up overwriting some French templates also from outside of wikitextprocessor. See [here](https://github.com/tatuylonen/wiktextract/blob/f00ef134673e842772a1852f701a3ab83c9dbef0/wiktextract/page.py#L3517-L3547).

The main nuisance is that from outside wikitextprocessor, there are no template objects and the strings (lines) have to be manipulated by indirect means (regex and such).

I processed a very small subset of the French Wikt dump with and without this change to wikitextprocessor and it only led to some minor sections not classified correctly. I would have to do more experiments to see if I can replicate the same functionality without this change to wikitextprocessor. Not sure when I will get around to do this since for my own needs the current solution works fine.

Another approach could be to inject template overwrites from the outside by adding an argument `template_overwrite` to the Wtp class that would take a function as an argument. Than the `template_overwrite` function could be called from within wikitextprocessor parsing `self.lang_code` and other necessary arguments to `template_overwrite`. However, I fear this makes everything even more obscure as it already is.

empiriker:
After careful examination, it turned out that `overwrite_fr_template` and [this logic](https://github.com/tatuylonen/wiktextract/blob/f00ef134673e842772a1852f701a3ab83c9dbef0/wiktextract/page.py#L3517-L3547) within `parse_page` from wiktextract are mostly redundant. This part in `parse_page` captures templates in a page's text with RegEx and replaces it before calling `ctx.parse` from wikitextprocessor.

`overwrite_fr_template` only had work to do when a template that should be overwritten wasn't correctly captured by a RegEx. I have improved the regular expressions (by allowing additional whitespace) and removed `overwrite_fr_template`.

If you agree with this approach, we can close this pull request.

__Considerations__
While this works for my case, I don't think it's the ideal solution. In essence, I have to redo part of the work of wikitextprocessor by crafting the correct RegExs to capture and replace Wikitext templates. As seen in my example, this is highly error-prone and would be avoidable by directly working on the parse tree that wikitextprocessor provides. 

If the goal of wikitextprocessor is to truly provide a "[Wikitext](https://en.wikipedia.org/wiki/Help:Wikitext) syntax parser that converts the whole page into a parse tree", there should be a way to manipulate this parse tree directly without having to change the source code of this package.

kristian-clausal:
It is not really wikitextprocessor's problem that the output we get from parsing French wiktionary does not work with wiktextract. The parse tree that wiktextract receives from wikitextprocessor is just a bunch of nodes, and manipulating those nodes is still possible; you can manually move, delete and create news nodes through assignments and manipulating node.children's content. Wiktextract can also call on wikitextprocessor to revert the parse tree back into wikitext, and then wiktextract can manipulate that wikitext and reparse it again if needed. Manipulating the wikitext source before parsing also works.

(If a wiktionary has some kind of novel syntax thing, then that's wikitextprocessor's problem, but not the structure or semantics of it.)

It's not elegant, yeah. But we are trying to extract data with a hand-written extractor based on input created by thousands of users over tens of different projects, each one with their own little mind and standards. 90% of wiktextract is just kludges to make it work halfway, because in reality, there is no single target. There is only kludge.

PR #36, opened 2023-04-25
Migrate to Lua 5.1

xxyzz:
The changes include:

- Migrate the legacy build script `setup.py` to `pyproject.toml`
- Migrate Lua code to Lua 5.1.
- Build lupa and lru-dict from source.


xxyzz:
I guess it's bug in older pip but fixed in later versions that the `lupa.lua51` module can't be found.

kristian-clausal:
Will be merged at the same as https://github.com/tatuylonen/wiktextract/pull/239, as long as https://github.com/scoder/lupa/pull/238 is merged so that we don't have to deal with a temporary fork of lupa.

xxyzz:
My Lupa pull request was merged, but now we have to use `pip install --use-pep517` option to build Lupa from source.

kristian-clausal:
Excellent! I have problems actually testing this without first merging it (because there are changes in both repos and if I try to `pip install` wiktextract it will just pull the old wiktextprocessor version because it hasn't been merged yet etc.) so I will merge now and if things don't work out, we can just revert the merges.

Today is Friday, so hopefully things work out and I don't get the usual "Kaikki.org failed to generate" e-mails on the weekend (as usually happens when any changes are made on a Friday).

xxyzz:
Please don't forget to use the `--use-pep517` pip install option. Should I create new pull request from the sqlite branches now?

kristian-clausal:
I will edit the regeneration script in the kaikki webgen repository, thanks for the reminder.

Tested installing wiktextract locally, and everything seemed to install and tests passed, and the files in site-packages seem to be correct source files. I will test a couple of pages and maybe a small run, let's see how it goes.

kristian-clausal:
Do you require --use-pep517 when installing wiktextract or wikitextprocessor? Not using or using it seems to result in the same installation. I think having `--use-pep517` in the setup for just Lupa seems to work correctly.

xxyzz:
Without `--use-pep517` pip will print some warning messages that suggest user to enable this option to install build dependency when building Lupa from source.

PR #38, opened 2023-05-02
make analyzing templates optional in `process_dump`

jmviz:
This is useful if one is not doing a full extract (i.e. phase1 and phase2), but instead calling `process_dump` directly, only caring about a certain page or module. In this case, `analyze_templates` just wastes time.

### DIFF
--- a/wikitextprocessor/dumpparser.py
+++ b/wikitextprocessor/dumpparser.py
@@ -202,7 +202,7 @@ def process_input(path, page_cb):
     return lst
 
 
-def process_dump(ctx, path):
+def process_dump(ctx, path, analyze_templates=True):
     """Parses a WikiMedia dump file ``path`` (which should point to a
     "<project>-<date>-pages-articles.xml.bz2" file.  This implements
     the first phase of processing a dump - copying it to a temporary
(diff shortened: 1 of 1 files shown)


kristian-clausal:
It is my understanding that `analyze_templates` is needed for the correct parsing of pages, because it generates data that the parser needs to know about which templates should be expanded first (because they create stuff that affects how the page is parsed in general, like lists etc.). I'll ask Tatu about it when I see him.

jmviz:
Ah, yes you're quite right. I was doing this while processing different language dumps to only extract their modules, so wasn't running into the problems with templates in wikitext that this would produce. I'll close this and make a different PR.

FYI, running `process_dump` on `cswiktionary-20230420-pages-articles.xml.bz2` hangs forever on `analyze_templates`. I discovered this while doing phase1 dump processing for recent dumps for every language that has data in `wiktextract/wiktextract/data`, under the assumption that this must have been successful for someone at some point. But now I am realizing that perhaps some of these language data have been added preemptively without anyone having tried to extract the corresponding dump. Either that or there has been some breaking change in `cs` wiktionary since whenever it previously worked.

kristian-clausal:
I asked Tatu about this, and he thought that the analyzed data could be find the wikt-cache data; so the analyzing actually doesn't seem necessary after all, we've just let it be because it can used with single pages for debugging (and it's quick enough on the dev machines that you don't really notice a second or two). However, with the changes coming now that wikt-cache is being migrated to a database, it's not productive to poke it at it right now.

PR #40, opened 2023-05-03
allow custom page_handler arg in process_dump

jmviz:
This allows custom treatment of individual pages based on their `model`, `title`, and `text`. For example, to process only modules, one could do:

```python
def page_handler(model, title, text):
        if title.startswith("Module:"):
            ctx.add_page(model, title, text)

ctx = Wtp()
process_dump(ctx, dump_file, page_handler=page_handler)


kristian-clausal:
Needs a bit more documentation, but looks fine.

PR #41, opened 2023-05-15
Save page texts to SQLite database

xxyzz:
Changes in this pull request:

- Decompress dump file with lbzcat if it's installed.
- Use lxml to parse the Wiktionary dump XML.
- Only save certain namespace pages.
- Save age texts to a SQLite database.


kristian-clausal:
I went through the diff and didn't notice anything, and learned a few tricks along the way. Only thing is formatting issues, the usual "80 characters per line" request and type-hinted function signatures are hard to read and could use a lot more line-splitting. Newlines don't cost anything, so being liberal with their use isn't an issue, there's plenty of that where it came from.

PR #48, opened 2023-05-18
Save extarcted pages to files

### LINKED ISSUE #47: Pages folder missing Modules and Templates, and issues with colons in word titles
Currently pages/ is missing Module/ and Template/ used for debugging, and possibly other folders I haven't used myself but could be useful. It's handy for grepping and for copying temp override files.

Additionally, two words [Swe:gë’](https://en.wiktionary.org/wiki/Swe:g%C3%AB%E2%80%99) and [Ohi:yoʼ](https://en.wiktionary.org/wiki/Ohi:yo%CA%BC) get misinterpreted as a Namespace:word pair, most probably due to the capitalization caused by their... proper nounedness. They have their own top-level Ohi and Swe folders.

xxyzz:
`Wtp.reprocess()` filters unused pages. In order to save all extracted pages, run the save files code inside `process_dump()`. Fixes #47 

Most changes to the `core.py` are formatting code made by [black](https://github.com/psf/black).

Opening and writing millions of files are slow, it takes ten minutes to create extracted pages. I'd discourage the usage of the `--pages-dir` option, page text can be inspected by querying the database.

Also fixes these Lua errors:
- https://kaikki.org/dictionary/errors/details-module-not-found-----RQ-xum-TI----export---.html
- https://kaikki.org/dictionary/errors/details-With-part-of-speech--ger---must-specif-Q9nKZHOa.html and other `Module:it-form of` errors.
- https://kaikki.org/dictionary/errors/details-attempt-to-concatenate-local--link---a-3wtv07QZ.html#LUA-error-in--invoke---coinage----coinage---parent---Template-coin----1---en---2---Q131761----in----1808---

### DIFF
--- a/wikitextprocessor/lua/_sandbox_phase2.lua
+++ b/wikitextprocessor/lua/_sandbox_phase2.lua
@@ -20,6 +20,8 @@ mw_python_get_page_info = nil
 mw_python_get_page_content = nil
 mw_python_fetch_language_name = nil
 mw_python_fetch_language_names = nil
+mw_wikibase_getlabel_python = nil
+mw_wikibase_getdesc_python = nil
 
 -- These are used for passing information about the current call to
 -- _lua_invoke().  The values are restred on return, as calls can be
@@ -195,10 +197,11 @@ end
 -- Python function that will be used for loading Lua modules and various
 -- other Python functions that implement some of the functionality needed
 -- for executing Scribunto code (these functions are called from Lua code).
-local function _lua_set_functions(mw_text_decode, mw_text_encode,
-                                  mw_text_jsonencode, mw_text_jsondecode,
-                                  get_page_info, get_page_content,
-                                  fetch_language_name, fetch_language_names)
+local function _lua_set_functions(
+        mw_text_decode, mw_text_encode, mw_text_jsonencode, mw_text_jsondecode,
+        get_page_info, get_page_content, fetch_language_name,
+        fetch_language_names, mw_wikibase_getlabel, mw_wikibase_getdesc
+)
    -- Note: this is exposed to the Lua sandbox and the Lua sandbox can access
    -- the functions via mw.  Thus all the Python functions provided here
    -- must be safe to call from hostile code.
@@ -211,6 +214,8 @@ local function _lua_set_functions(mw_text_decode, mw_text_encode,
    mw_python_get_page_content = get_page_content
    mw_python_fetch_language_name = fetch_language_name
    mw_python_fetch_language_names = fetch_language_names
+   mw_wikibase_getlabel_python = mw_wikibase_getlabel
+   mw_wikibase_getdesc_python = mw_wikibase_getdesc
 end
 
 
--- a/wikitextprocessor/lua/_sandbox_phase2.lua
+++ b/wikitextprocessor/lua/_sandbox_phase2.lua
@@ -34,7 +34,13 @@ local function frame_args_index(new_args, key)
       key = i
    end
    local v = new_args._orig[key]
-   if v == nil then return nil end
+   if v == nil then
+       -- "inf" got converted to number type, convert back to string
+       v = new_args._orig[tostring(key)]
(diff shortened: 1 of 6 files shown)


wikitextprocessor/luaexec.py
```diff
@@ -574,22 +594,66 @@ def expandTemplate(frame, *args):
...
+        '<strong class="error">{} in Module:{} function {}'
+        "</strong>".format(msg, html.escape(modname), html.escape(modfn))
+    )
+
+
+@functools.cache
+def query_wikidata(item_id: str):
+    import requests
+
+    r = requests.get(
```
kristian-clausal, line 632: We don't query data in other places, so this is a little out of place. Ideally, we should just gather the data cached in this function, make it a separate dict or json file, and load the data from there, updating as needed.

We could do this by using an error/debug/warning message each time 'query_stored_wikidata' is called and it can't find the value in the previously stored data. The error messages can be extracted by a separate script that can then reimplement the functionality here to query data (using the old keys and error keys) and save it locally, until we get more "missing wikidata" errors. I also think all of this should probably happen in Wiktextract, it feels like most things having to do with data should go there and only passed on to wikitextprocessor.

Currently, this should work (afaict), so the above is for the future.

wikitextprocessor/lua/_sandbox_phase2.lua
```diff
@@ -34,7 +34,13 @@ local function frame_args_index(new_args, key)
       key = i
    end
    local v = new_args._orig[key]
-   if v == nil then return nil end
+   if v == nil then
+       -- "inf" got converted to number type, convert back to string
```
kristian-clausal, line 40: Could this error happen with names like "ind" (<- #IND?) or "nan"? Seems like this should fix all of them. This bug is *weird*!

wikitextprocessor/luaexec.py
xxyzz: The API results are cached(please see the above `@functools.cache` line).

kristian-clausal:
Seems good to me, although the wikidata query should be extracted at some point so that the data is provided as a local file through wiktextract.

xxyzz:
But there are countless Wikidata items(like Q:12345), it'd be impractical to save them to a JSON file.

kristian-clausal:
We only need to save the data that is actually used by Wiktionary.

wikitextprocessor/luaexec.py
kristian-clausal: Yes, the results are cached when wikitextprocessor is running, but we probably do not want to query wikidata each time wikitextprocessor runs. We should pass data into wikitextprocessor from wiktextract (and most preferably local data without querying wikidata each time).

xxyzz:
Tracking that many data manually in text file with git is not a good idea... We already have a few MB size files.

wikitextprocessor/lua/_sandbox_phase2.lua
xxyzz: I have not idea how "inf" got converted to a number...

kristian-clausal: Clarifying comment for people reading this in the future:
Apparently in (current?) Lua, you can use #INF and #IND to represent infinity (and I forget what #IND was), and the to_string return for the float value infinity is "inf" or "-inf" (and for NaN it's "nan", "-nan"). Maybe the name is evaluated as an expression (for some reason?), which turns it into a float somewhere?

xxyzz: But Lua 5.1 doesn't have the `inf` variable, `loadstring("inf")` returns `nil`. Either `inf` was defined somewhere or `key` was converted to number by `tonumber()` before passing to `frame_args_index()`.

xxyzz: Damn, tonumber("inf") returns the number `inf`.

PR #50, opened 2023-05-22
Match URLs in template parameter which contain whitespaces at the start

xxyzz:
This is a follow up fix for the commit https://github.com/tatuylonen/wikitextprocessor/commit/dfb34470424562437c159220c362cf5843da82b8.

Some pages add whitespace before the template parameter value, for example: `{{ template_name | url = https://github.com }}`. And whitespaces are stripped from the beginnings and ends of named parameter names and values: https://en.wikipedia.org/wiki/Help:Template#Parameters


xxyzz:
I was planning to fix the `Only one of date= or year= should be specified` Lua error in the ["went"](https://en.wiktionary.org/wiki/went) page. But this error looks legit to me as the second unnamed parameter passed as `year` parameter, I don't know how mediawiki ignores this error.

PR #51, opened 2023-05-23
Don't skip whitespace token in `parser.token_iter()`

xxyzz:
Previous commit eecb07352bc7eebc4924ed08d24b1d3fedce236a incorrectly added these two lines. This would cause the parser deletes the space between words. For example, page https://en.wiktionary.org/wiki/丫鬟 uses a `c=aa1 waan4` template parameter, but remove the whitespace would cause error in https://en.wiktionary.org/wiki/Module:yue-pron because it checks if the parameter matches "[0-9][a-z]" pattern.


kristian-clausal:
Thanks for this fix, I was looking at #252 too.

PR #53, opened 2023-05-25
Don't commit changes to db if some pages are temporally overwritten

xxyzz:
Also only run `Wtp.analyzed_templates()` if it's the first time parsing a dump file or template pages are overwritten or the results of `analyzed_templates()` are not saved to db last time.

### DIFF
--- a/tests/test_wikiprocess.py
+++ b/tests/test_wikiprocess.py
@@ -6,8 +6,8 @@ import os
 import math
 import time
 import unittest
-import platform
 
+from pathlib import Path
 from typing import List, Tuple
 
 from wikitextprocessor import Wtp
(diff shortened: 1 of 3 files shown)


xxyzz:
Even tough all tests are passed I'm not sure if this could work correctly with multiprocessing without committing changes to database.

xxyzz:
Backup database to another file then use this file next time is simpler and more reliable. This could avoid "database locked" error in my first commit.

kristian-clausal:
While reading about SQLite temporary databases, my understanding is that a SQLite temporary database is actually a temporary physical file. How costly do you think it would be to copy the database into the temporary database and overwrite that temporary database?

xxyzz:
Create a backup db file takes about 4 seconds. Here are the test code and result:

```python
import sqlite3

source = sqlite3.connect("sqlite_db")
dest = sqlite3.connect("backup_db")
with dest:
    source.backup(dest)
dest.close()
source.close()
```

```
$ time python backuptime.py
real    0m4.626s
user    0m0.904s
sys     0m3.557s
```

I still prefer updating the database directly with a Python script or [DB Browser for SQLite](https://sqlitebrowser.org) while debug the code.

kristian-clausal:
That's amazingly fast!

xxyzz:
Except that if an overwrite page is a template page, then `Wtp.analyze_template()` will be called. But if you know the change won't affect the analyze result you can modify the page content on the database and skip the `analyze_template()`.

PR #55, opened 2023-05-29
Clear `mw.loadData()` cache before processing each page

xxyzz:
According to Scribunto
document(https://www.mediawiki.org/wiki/Extension:Scribunto/Lua_reference_manual#mw.loadData), `mw.loadData()` evalutates the loaded module only once per page, rather than once per `{{#invoke:}}` call. The cache should be cleared for each page, otherwise some module like
`Module:headword/data`(https://en.wiktionary.org/wiki/Module:headword/data) would use old cached page title.

### DIFF
--- a/wikitextprocessor/core.py
+++ b/wikitextprocessor/core.py
@@ -103,7 +103,8 @@ class Wtp:
         "lua",  # Lua runtime or None if not yet initialized
         "lua_depth",  # Recursion depth in Lua calls
         "lua_invoke",  # Lua function used to invoke a Lua module
-        "lua_reset_env",  # Function to reset Lua environment
+        "lua_reset_env",  # Lua function to reset Lua environment
+        "lua_clear_loaddata_cache",  # Lua function to clear mw.loadData() cache
         "lua_path",  # Path to Lua modules
         "num_threads",  # Number of parallel threads to use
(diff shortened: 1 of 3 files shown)


kristian-clausal:
Good morning to you and good morning to *me*! This fix looks excellent to me, I suspect this issue might be the source of many of our Lua troubles.

xxyzz:
Morning! I also add support of multiple pages to the `--page` command option to debug this error, I'll create a pull request later.

kristian-clausal:
> Morning! I also add support of multiple pages to the `--page` command option to debug this error, I'll create a pull request later.

That would be great, it's been on my wish-list for a while, thanks!

kristian-clausal:
Testing out an individual page (Japanese stuff, testing out cleaning up stuff related to `package`) there feels like a slowdown. I'll run the kaikki.org regen process starting now, about seven, eight hours in advance of the default start of a run, so we'll see if there's any major slow downs.

xxyzz:
But this change won't affect the speed when debugging a single page since the cache Lua table is cleared only before processing each page. I guess it could be slower when processing the entire dump file.

kristian-clausal:
It might just be in my head, or the specific page I was testing.

Issue #63, opened 2023-06-20
Filenames cause errors on exFAT


brendanedwardgavin:
exFAT partitions disallow file names with the characters `/\:*?\"<>|` and unicode 0x00-0x1F; some wiktionary titles violate this rule, causing the following error when extracting the wiktionary raw dump:
```
2023-06-20 13:26:17,038 INFO:   ... 8420000 raw pages collected
Getting all pages for query: 'SELECT title, namespace_id, redirect_to, need_pre_expand, body, model FROM pages ORDER BY title ASC'
Traceback (most recent call last):
  File "/home/brendan/miniconda3/envs/wiktextract/bin/wiktwords", line 8, in <module>
    sys.exit(main())
  File "/home/brendan/miniconda3/envs/wiktextract/lib/python3.10/site-packages/wiktextract/wiktwords.py", line 331, in main
    parse_wiktionary(wxr, args.path, word_cb,
  File "/home/brendan/miniconda3/envs/wiktextract/lib/python3.10/site-packages/wiktextract/wiktionary.py", line 94, in parse_wiktionary
    for _ in wxr.wtp.process(
  File "/home/brendan/miniconda3/envs/wiktextract/lib/python3.10/site-packages/wikitextprocessor/core.py", line 1513, in process
    process_dump(
  File "/home/brendan/miniconda3/envs/wiktextract/lib/python3.10/site-packages/wikitextprocessor/dumpparser.py", line 120, in process_dump
    save_pages_to_file(ctx, save_pages_path)
  File "/home/brendan/miniconda3/envs/wiktextract/lib/python3.10/site-packages/wikitextprocessor/dumpparser.py", line 201, in save_pages_to_file
    file_path.parent.mkdir(parents=True, exist_ok=True)
  File "/home/brendan/miniconda3/envs/wiktextract/lib/python3.10/pathlib.py", line 1175, in mkdir
    self._accessor.mkdir(self, mode)
OSError: [Errno 22] Invalid argument: '../wiktextract-dump-2023-06-20/Words/!?'
```

kristian-clausal:
[...]

PR #64, opened 2023-06-20
Fix exFAT filename error

brendanedwardgavin:
Resolves Issue #63

### DIFF
--- a/wikitextprocessor/dumpparser.py
+++ b/wikitextprocessor/dumpparser.py
@@ -187,11 +187,11 @@ def path_is_on_windows_partition(path: Path) -> bool:
     Return True if the path is on an exFAT or NTFS partition.
     """
     path_matching_fstypes_mountpoints = [
-        (part.fstype, part.mountpoint) for part in disk_partitions() 
+        part for part in disk_partitions() 
         if str(path.resolve()).startswith(part.mountpoint)
     ]
     #we want the more specific (i.e. longer) matching mountpoint
-    path_fstype = sorted(path_matching_fstypes_mountpoints, key=lambda x: len(x[1]))[-1][0].lower()
+    path_fstype = sorted(path_matching_fstypes_mountpoints, key=lambda x: len(x.mountpoint))[-1].fstype.lower()
     return path_fstype == "exfat" or path_fstype == "fuseblk" or path_fstype == "ntfs"
 
 
--- a/wikitextprocessor/dumpparser.py
+++ b/wikitextprocessor/dumpparser.py
@@ -12,13 +12,27 @@ import sys
 
 from collections.abc import Callable
 from pathlib import Path
-from typing import Optional, Set, List, IO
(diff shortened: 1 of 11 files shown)


kristian-clausal:
I'll ok this if you could put the exFAT stuff behind a command-line option like `--exfat-safe`.

wikitextprocessor/dumpparser.py
```diff
@@ -5,13 +5,15 @@
 import hashlib
 import logging
 import os
+from psutil import disk_partitions
```
xxyzz, line 8: This would get import error unless package `psutil` is added as dependency to the` pyproject.toml` file.

wikitextprocessor/dumpparser.py
```diff
@@ -180,12 +182,44 @@ def overwrite_pages(
...
+def path_is_on_windows_partition(path: Path) -> bool:
+    """
+    Return True if the path is on an exFAT or NTFS partition.
+    """
+    path_matching_fstypes = [
+        part.fstype for part in disk_partitions() 
+        if str(path.resolve()).startswith(part.mountpoint)
+    ]
+    path_fstype = sorted(path_matching_fstypes, key=len)[-1]
+    return path_fstype == "exfat" or path_fstype == "fuseblk"
```
xxyzz, line 194: Why only check the last matching file system? And I think "fuseblk" is what NTFS partition displays on Linux but would certainly shows a different value on Windows, maybe just "ntfs".

kristian-clausal: It's looking for a best match in case you've got nested mounts based on this https://stackoverflow.com/a/35291824

wikitextprocessor/dumpparser.py
kristian-clausal: Adding psutil is as a dependency is not optimal, but should be fine... I'm surprised the Python standard libraries don't have something for this.

wikitextprocessor/dumpparser.py
xxyzz: That's different, here the `path_fstype` variable comes from a list of file systems types sorted by string length not mount point path.

kristian-clausal: You're correct, I just saw the post just beneath that (by bgav) and assumed. Programming is hard, debugging is even harder.

brendanedwardgavin: @xxyzz @kristian-clausal thanks for catching those mistakes, I've updated the PR to fix them. I also added tests, which really I should have done the first time around. I thought this would be simple but clearly I was wrong.

xxyzz:
LGTM.

wikitextprocessor/dumpparser.py
kristian-clausal: > I thought this would be simple but clearly I was wrong.

I see you've read our motto.

kristian-clausal:
Thanks!

PR #65, opened 2023-06-26
pop parser stack to the correct list node

xxyzz:
If the list token is "#:", then the code pops stack until the stack
top is the second list in the stack.

This fixes the nested list in English Wiktionary page "a"'s "Norwegian
Bokmål" section:

```
# the letter [[#English|a]], the first letter of the Norwegian alphabet
#* {{quote-lite|nb|year=1999|author={{w|Lars Roar Langslet}}|title=I kamp for norsk kultur|page=234|passage=bruken av a i bestemt form i hunkjønnsord|t=the use of a in the definite form of feminine words}}
#: {{ux-lite|nb|[[fra A til B]]|[[from A to B]]|inline=1}}
#: {{ux-lite|nb|[[fra A til Å]]|[[from A to Z]]|inline=1}}
```

I can't add this WikiText to git commit message, it's as if git or my text editor sanitizes the code.

Template "quote-lite" expands to HTML tags but some tags are unclosed,
and it has two newlines at the end which pop the stack too early for
the following example lists. Therefore don't pop list nodes if the
token is "\n".

And new `pop_until_nth_list` function is similar to the previous code in the `while` loop block in the `list_fn` function, but I think leave the loop unchanged at the moment would be safer.

Fixes https://github.com/tatuylonen/wiktextract/issues/267.

### DIFF
--- a/wikitextprocessor/parser.py
+++ b/wikitextprocessor/parser.py
@@ -384,7 +384,8 @@ def _parser_pop(ctx, warn_unclosed):
     # chilren list.
     ctx.parser_stack.pop()
 
-def _parser_have(ctx, kind):
+
+def _parser_have(ctx: "Wtp", kind: NodeKind) -> bool:
     """Returns True if any node on the stack is of the given kind."""
     assert isinstance(kind, NodeKind)
     for node in ctx.parser_stack:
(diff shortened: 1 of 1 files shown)


xxyzz:
The test GitHub Actions somehow doesn't run for this pull request, but it should be running. Here is the test from my forked repo: https://github.com/xxyzz/wikitextprocessor/actions/runs/5375654529

kristian-clausal:
This seems fine and I'll merge it, but I want to try out another way to pop up the list hierarchy to see how well it works.

xxyzz:
There is another way? Do you mean your comment at here: https://github.com/xxyzz/wikitextprocessor/commit/5df3a172aa42a68340eda6892b4deffa9dcbdd2c#commitcomment-119633559?

but that sounds similar to this pull request's implementation.

kristian-clausal:
I think you can do it without counting the layers with a simple comparison. The list has a string argument (`LIST #*` in your tree print in the commit) that can be compared to other lists to see whether it's higher or lower in the hierarchy with string comparisons. I'll see if I can get it to work or if I'm wrong.

xxyzz:
I think that's would be unnecessary... the node order in the stack already represent their hierarchy level and I think you also need to iterate the stack and check it's a list type node besides the string argument check.

PR #68, opened 2023-06-30
Check template body is not `None` explicitly in `Wtp.expand.expand_recurse()`

xxyzz:
`if body` is also `False` if variable `body` is an empty string, thus the variable `t` will remain `None` and later the `assert isinstance(t, str)` code will fail. Actually, `body` can't be `None`, because the template name is check with `Wtp.template_exists()` before.

Example template: https://zh.wiktionary.org/wiki/Template:中2 this template's body is empty string after proccessing with `Wtp.__template_to_body()`.


kristian-clausal:
Thank you, I was just about to delve into this because Kaikki's regeneration failed with exceptions that (I see now) are related to this.

Issue #59, opened 2023-06-15
Template regex in `Wtp._encode()` can't match templates that has `-{}-` argument


xxyzz:
Chinese Wiktionary template [Ja-romanization of](https://zh.wiktionary.org/wiki/Template:Ja-romanization_of) uses `-{}-` as a placeholder for the first unnamed argument of module [form of/templates](https://zh.wiktionary.org/wiki/Module:Form_of/templates), `-{}-` will be replaced to empty string by MediaWiki. But the template regex at here https://github.com/tatuylonen/wikitextprocessor/blob/04cec9ff80094c91c8174b7dab125629399c081c/wikitextprocessor/core.py#L563-L572

can't match the `#invoke` function and the whole invoke function got expanded as plain text.

`ja-romanization of` template wikitext: `{{#invoke:form of/templates|form_of_t|-{}-|withcap=1|lang=ja|noprimaryentrycat=}} 的[[罗马字]]转写`

I tried to replace `-{}-` with a white space before calling `_encode` but I guess some Lua code removes the empty string argument and throws a "parameter 1 is required" error.  Maybe a more complex regex could solve this bug.

Example of affected page: https://zh.wiktionary.org/wiki/manga#日語

kristian-clausal:
We could steal a MAGIC character slot, replace `-{}-` with it, and then return it as needed.

In common we have:
https://github.com/tatuylonen/wikitextprocessor/blob/04cec9ff80094c91c8174b7dab125629399c081c/wikitextprocessor/common.py#LL8C1-L18C1

The magic characters and the range are just unused Unicode characters that should be safe to sprinkle into strings.

Here we could add a MAGIC_ZH_PLACEHOLDER by shifting MAGIC_FIRST to 0x0x00102040 and giving 0x0010203f to MAGIC_ZH_PLACEHOLDER, cast it as a character in MAGIC_ZH_PLACEHOLDER_CH and do string replacements back and forth between the magic character and `-{}-` as needed.

(Note to self: replace magic values with MAGIC_BASE_VALUE + N.)

This is what I did when I needed to somehow fix an issue with quote single quote `'` apostrophes inside double-quoted HTML attributes when parsing.

PR #73, opened 2023-07-05
Update template matching regex in `Wtp._encode()`

### LINKED ISSUE #72: Paser can't parse templates that have `{}` in arguemnt
Similar to #59, example page: https://zh.wiktionary.org/wiki/%, wikitext:

```
#: {{zh-x|約 有 6 '''%'''{pā} 的 臺灣人 血型 是 A{ēi}B{bī}型。|}}
```

The text inside curly brackets is Pinyin for the previous character, this data is ignored by MediWiki and not displayed in the expanded HTML. I haven't find the Wikitext document for this syntax, maybe it's already obsolete. This time a temporary fix with magic character would be hard to implement.

xxyzz:
Fixes #72 and #59

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -20,6 +20,7 @@ def parse_with_ctx(title, text, **kwargs):
 
 def parse(title, text, **kwargs):
     root, ctx = parse_with_ctx(title, text, **kwargs)
+    ctx.close_db_conn()
     assert isinstance(root, WikiNode)
     assert isinstance(ctx, Wtp)
     return root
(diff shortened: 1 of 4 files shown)


kristian-clausal:
I wish I knew what was happening in that regex. LGTM!

xxyzz:
I use this [website](https://regex101.com/r/BMxRFB/1) to see which part of the text is matched but I still don't know what kind of text the first two alternatives regex patterns(`{\|[^{}]*?\|}` and `}[^{}]`) are trying to match.

kristian-clausal:
The first pattern is the outermost brackets of a table, but the second one is still weird to me. I'm going through the section with this, and it could really need with some documentation and maybe some named consts.

xxyzz:
A table as a template argument? Could you find an example?

The second alternative pattern was added in https://github.com/tatuylonen/wikitextprocessor/commit/ee043cff190543fb94cb40d4827444d8982a30fe, looks like the second pattern was added by mistake.

kristian-clausal:
Wikitext documentation seems to indicate you should be able to add them. It's hard to test this because I need to find a template that just... prints out stuff.

Trying with #if, I can't get it to accept {|.
```
{{#if: true|
{{{!}}
{{!}}-
! tesssst !! tesst !! tesssssssst
{{!}}-
{{!}}test {{!}}{{!}} test2 {{!}}{{!}} test3
{{!}}}
|}}
```
Have to escape the vbars with `{{!}}`.

xxyzz:
This first `{\|[^{}]*?\|}` pattern was added from this commit: https://github.com/tatuylonen/wikitextprocessor/commit/1e7539728e25e4f35c20467d36916b6b5fc3771a#diff-9d58f4d4915c12ad491e1b91e228f1033209f5a4b0078dd6d68f7f39dd0d2a87R345

The commit message says the change was for parsing incorrect Wikitext.

PR #74, opened 2023-07-05
Fix catastrophic backtracking in `Wtp._encode()`'s template regex

xxyzz:
previous regex halts code on many pages, example page: https://zh.wiktionary.org/wiki/路用

I really should read a regex book now...

### DIFF
--- a/wikitextprocessor/core.py
+++ b/wikitextprocessor/core.py
@@ -662,12 +662,12 @@ class Wtp:
                     break
             # Replace template invocation
             text = re.sub(
-                r"{" + MAGIC_NOWIKI_CHAR + r"?{(("
+                r"{" + MAGIC_NOWIKI_CHAR + r"?{((?:"
                 r"{\|[^{}]*?\|}|"  # Outer table tokens
                 r"}[^{}]|" # lone }????
                 r"[^{}](?:{[^{}|])?|"  # lone possible {???
                 r"-{}-|"  # GitHub issue #59 Chinese wiktionary special `-{}-`
-                r"[^{}]+{[^{}]+}"  # GitHub issue #72 zh.wiktionary `A{pron}`
+                r"[^{}]{[^{}]+}"  # GitHub issue #72 zh.wiktionary `A{pron}`
                 r")+?)}" + MAGIC_NOWIKI_CHAR + r"?}",
                 repl_templ,
                 text,


wikitextprocessor/core.py
```diff
@@ -662,12 +662,12 @@ def repl_extlink(m):
...
             # Replace template invocation
             text = re.sub(
-                r"{" + MAGIC_NOWIKI_CHAR + r"?{(("
+                r"{" + MAGIC_NOWIKI_CHAR + r"?{((?:"
                 r"{\|[^{}]*?\|}|"  # Outer table tokens
                 r"}[^{}]|" # lone }????
                 r"[^{}](?:{[^{}|])?|"  # lone possible {???
                 r"-{}-|"  # GitHub issue #59 Chinese wiktionary special `-{}-`
-                r"[^{}]+{[^{}]+}"  # GitHub issue #72 zh.wiktionary `A{pron}`
+                r"[^{}]{[^{}]+}"  # GitHub issue #72 zh.wiktionary `A{pron}`
```
kristian-clausal, line 670: This might be triggered by `A{aa}}`, so might need to add something like `[^}]` at the end, but then it won't catch `{{A{aa}}}`...

xxyzz: Maybe the regex could be simplified to `{{((?:[^{}]|-{}-|{[^{}]+})+?)}}` but I haven't test it on the dump file, all the unit tests are passed.

kristian-clausal: Tatu arrived an hour ago and we've been looking at the regex, I'll be simplifying some of the regex, but he was puzzled that the regex from before issue #72 didn't catch the curly braces. That's what the two mystery lines are for:

```
                r"}[^{}]|" # lone }????
                r"[^{}](?:{[^{}|])?|"  # lone possible {???
```
These should catch any lone curly braces, which the ones in `a{aa}` should be. I'll test this out with a regex tester.

kristian-clausal: On https://regex101.com/ (switch to the Python engine), `{{(({\|[^{}]*?\|}|}[^{}]|[^{}](?:{[^{}|])?)+?)}}` is able to pick up `{{foo a{a} }}`; does this work with zh-pages? I left out the "-{}-" because it wasn't relevant, but it needs to be part of the regex as a special case.

kristian-clausal: Tatu also wished for the curly braces to be escaped, just in case.

kristian-clausal: > On https://regex101.com/ (switch to the Python engine), `{{(({\|[^{}]*?\|}|}[^{}]|[^{}](?:{[^{}|])?)+?)}}` is able to pick up `{{foo a{a} }}`; does this work with zh-pages? I left out the "-{}-" because it wasn't relevant, but it needs to be part of the regex as a special case.

This doesn't work with `{{zh-x|約 有 6 '''%'''{pā} 的 臺灣人 血型 是 A{ēi}B{bī}型。|}}`! Weird!

kristian-clausal: It works if you remove parts of the `{bī}`. An issue with regex consumption.

kristian-clausal: The issue was with B ovelapping in the above case, which could be fixed with a regex lookahead (had to look that up, and yep, modern Python has regex lookahead): `{{(({\|[^{}]*?\|}|}(?=[^{}])|[^{}]{[^{}|]|[^}{])+?)}}`, `(?=...)` look for `...` ahead, but do not consume. It works with the example from above.

kristian-clausal:
I will merge this for the tests, and use the regex with the lookahead; hopefully it is correct. It was able to find the innermost templates in the test text, and then when those innermost templates were replaced with `X` (magic char cookie), it was able to match the whole example.

xxyzz:
Why not use a simpler `{{((?:[^{}]|{[^{}]*})+)}}`?

And I run this SQL `SELECT title FROM pages WHERE body REGEXP '\{\{[^{}]*\{\|[^{}]*\|\}[^{}]*\}\}';` to see if any page actually has a table template argument but this SQL only find page [Template:no-verb overview](https://en.wiktionary.org/wiki/Template:no-verb_overview) and it's template parameter(`{{{|past-part-bracket2|}}}`) not a table and it's not used in dictionary entry page.

IMO, escape curly brackets would be a little too cautious and make the regex harder to read.

kristian-clausal:
We are currently matching singleton `{` and `}`, and I believe that is because Tatu came across them. I didn't specifically ask whether they were part of pairs, but I believe it is correct to parse them, even if they're unbalanced. There could be, now or in the future, a template that just goes `{{foo| } | bar}}` and we should be prepared.

That `{{{|past-part-bracket2|}}}` is definitely a typo.

I will ask Tatu again about the table bracketting, it must have been a problem at some point, but it's not necessarily a problem anymore if Wiktionary or (even better) Wikitext has become stricter about it.

PR #78, opened 2023-07-12
Fix template name variable got overwritten by called template name bug

xxyzz:
`Wtp._analyze_template()`'s local variable for called template name overwrites the analyzed template name, which would cause the following check Chinese Wiktionary heading template code uses the called template name.


kristian-clausal:
I've been fixing a few reassignment type conflicts here and there, but this I wouldn't have even noticed.

PR #87, opened 2023-08-17
Separate WikiNode.args into WikiNode.sarg and WikiNode.largs

kristian-clausal:
WikiNode previously had one attribute that held two kinds of argument data: either a single string, for things like HTML tag names, List "##*"-style level strings etc., or a list of lists containing more complicated data, like Template arguments and stuff.

It's really hard to keep track of, and a really annoying thing to write well-typed code for. Because by necessity you'd need to have it be a [str | List[List[...]]], you have to constantly create phantom asserts either with `# type: ignore` or `if TYPE_CHECKING: assert...` to make type-checkers happy.

`sarg` and `largs` are both preliminary names: string argument and list arguments, specifically, but if someone can come up with something better, these two should be unique enough to be easily replaceable.

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -24,7 +24,7 @@ class ParserTests(unittest.TestCase):
         tree = self.parse("test", "")
         self.assertEqual(tree.kind, NodeKind.ROOT)
         self.assertEqual(tree.children, [])
-        self.assertEqual(tree.args, [["test"]])
+        self.assertEqual(tree.largs, [["test"]])
 
     def test_text(self):
         tree = self.parse("test", "some text")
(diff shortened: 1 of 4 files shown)


xxyzz:
IMO the problem is the `WikiNode` class is too general, rename the attribute just make the type checking easier but not make writing extractor code easier. You mentioned another approach before: create class inherent the `WikiNode` class, and the type and attribute name could be changed accordingly. Like a HTML node could be `HTMLNode` and has a `str` type `tag` attribute. And template node could have a `dict` type `arguments` attribute, a `str` type `template_name` attribute. But I'm not sure the implantation complicity. I'm not again the change though we could improve it later.

kristian-clausal:
It does feel like we could use a lot more classes than we do now, especially when using different WikiNode subclasses is almost free, because WikiNodes are already... classed objects, you know what I mean.

However, this is the smallest change I could think of that shouldn't break anything, and it will also make it easier to do later changes (`.sarg` and `.largs` are easier to look for than`.args` which already has overlap in a ton of Lua code). If there's no big issues, I think we can safely merge this change.

PR #88, opened 2023-08-18
Support multiprocessing on macOS and Windows

xxyzz:
I'm also planning to replace the deprecated `pkg_resources` module, I'll do that next week.


xxyzz:
https://github.com/tatuylonen/wikitextprocessor/pull/88/commits/4b3d9631f2339984abd94cddb506b3e3c5387a5a this makes the code 4 times faster. But I don't know why this simple SQL is slow.

wikitextprocessor/core.py
```diff
@@ -1683,6 +1684,7 @@ def magic_repl(m: re.Match) -> str:
         # print("    _finalize_expand:{!r}".format(text))
         return text
 
+    @lru_cache
```
kristian-clausal, line 1685: Tatu recommends setting a cache size here (the default is surprisingly small, 128, so about 4000 is good?) and that this cache is probably what caused the speed-up because zh.wiktionary has a lot of get_page calls. If * star in the SQL query doesn't contain other fields, then the latter query with explicit titles should be about exactly the same in speed.

xxyzz: I think it has more affects on Chinese dump file is because Chinese extractor code call `call_lua_sandbox` less than the English code and `get_page` is the bottleneck. I set the limit to 1000, probably should be enough. The default size already makes `get_page` take less than 1% of the time and reduce the extract time to 2 hours.

wikitextprocessor/core.py
```diff
@@ -973,9 +973,7 @@ def analyze_templates(self) -> None:
             "Analyzing which templates should be expanded before parsing"
         )
         # Add/overwrite templates
-        template_ns: NamespaceDataEntry = self.NAMESPACE_DATA.get(
-            "Template", EMPTY_NAMESPACEDATA
```
kristian-clausal: Please stop removing annotations when they are put there as documentation.

xxyzz: This type annotation is unnecessary because you already added the type to `NAMESPACE_DATA`.

kristian-clausal: It is there as a comment. Just leave them be. We have already talked about this, annotations are not just for the type checker, it also helps to comment the code.

xxyzz: OK, I restored it. I just feel this is kind obvious unlike some complicated cases and type checker get not problem to find the type so I deleted it.

kristian-clausal: Thanks. Named types are named for a reason, either they're complicated, like NamespaceDataEntry which is a pretty big TypedDict, or they're simple but it's helpful to know what paths stuff is supposed to take. A `str` is not necessarily just a `str`, and type annotations and variable names are helpful to telling us the *whys* and the *wheres*.

In general, if it's not a basic Python function that produces sometimes immediately obvious like a str or int, the named type annontations are helpful, and they never cause harm. They're also useful as sentinels to keep an eye out on expected values and actual values.

xxyzz: I usually won't bothered to add type to value returned from a typed function... Tools in text editors are pretty good for finding errors and code completion.

xxyzz:
Lua errors in Wiktionary Module pages won't throw `lupa.LuaError` exception because [`pcall`](https://www.lua.org/manual/5.1/manual.html#pdf-pcall) Lua api doesn't propagate errors. But if error happens in other part of the Lua code in `lua` folder, then https://github.com/tatuylonen/wikitextprocessor/pull/88/commits/a4a0bfc09f006cad68ccf6c220969f34fdf081c1 will be used.

PR #89, opened 2023-08-24
Add `TemplateNode` and `HTMLNode` subclasses

xxyzz:
The parse is broken for the two HTML tests...

### DIFF
--- a/wikitextprocessor/core.py
+++ b/wikitextprocessor/core.py
@@ -1535,10 +1535,11 @@ class Wtp:
                         # print("TEMPLATE_FN {}: {} {} -> {}"
                         #      .format(template_fn, name, ht, repr(t)))
                     if t is None:
-                        body: Optional[str] = self.read_by_title(
+                        template_page = self.get_page_resolve_redirect(
                             name, self.NAMESPACE_DATA["Template"]["id"]
                         )
-                        if body is not None:
+                        if template_page is not None:
+                            body = template_page.body
                             # XXX optimize by pre-encoding bodies during
                             # preprocessing
                             # (Each template is typically used many times)
@@ -1551,17 +1552,7 @@ class Wtp:
                             encoded_body = expand_args(encoded_body, ht)
                             # Expand the body using the calling template/page
                             # as the parent frame for any parserfn calls
-                            new_title = tname.strip()
-                            for prefix in self.NAMESPACE_DATA:
-                                if tname.startswith(prefix + ":"):
-                                    break
-                            else:
-                                new_title = (
-                                    self.NAMESPACE_DATA["Template"]["name"]
-                                    + ":"
-                                    + new_title
-                                )
-                            new_parent = (new_title, ht)
+                            new_parent = (template_page.title, ht)
                             # print("expanding template body for {} {}"
                             #       .format(name, ht))
                             # XXX no real need to expand here, it will expanded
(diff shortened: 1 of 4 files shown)


wikitextprocessor/parser.py
```diff
@@ -334,11 +334,40 @@ def __str__(self) -> str:
...
+        for node in self.find_child_recursively(target_kind):
+            return True
+        return False
 
 class TemplateNode(WikiNode):
     def __init__(self, linenum: int):
         super().__init__(NodeKind.TEMPLATE, linenum)
-        self._template_parameters: Optional[Dict[str | int, str | WikiNode]] = None
+        self._template_parameters: Optional[
+            Dict[str | int, str | WikiNode]
```
kristian-clausal, line 380: We haven't yet upgraded to a version of Python that support the pipe for type annotations. Please use Union[]

xxyzz: I forget to use Union... Actually we can use `|`, it was introduced in Python 3.10.

xxyzz:
> There are a few places in the code where the type of node a WikiNode has is changed through assigning a different NodeKind to it, just to keep an eye out.

I didn't know that, will search the code to see how the node type is changed.

And the html test returns strange tree, I'm not sure if this is expected:

The parse return of `<h1><p></p></h1>` is `<ROOT(['t']){} <HTML(h1){} >, <HTML(p){} >, '</h1>'>`, this doesn't look right.

kristian-clausal:
We have 3.8 here, so that's what we're targeting. The newest Python available on certain distros. On Aug 24, 2023 15:52, xxyzz ***@***.***> wrote:
@xxyzz commented on this pull request.



In wikitextprocessor/parser.py:
>  
 class TemplateNode(WikiNode):
     def __init__(self, linenum: int):
         super().__init__(NodeKind.TEMPLATE, linenum)
-        self._template_parameters: Optional[Dict[str | int, str | WikiNode]] = None
+        self._template_parameters: Optional[
+            Dict[str | int, str | WikiNode]

I forget to use Union... Actually we can use |, it was introduced in Python 3.10.

wikitextprocessor/parser.py
xxyzz: Maybe we could consider using [container](https://podman.io) or [pyenv](https://github.com/pyenv/pyenv) on the server, using Python 3.11 is the easiest way to improve speed: https://docs.python.org/3/whatsnew/3.11.html#faster-cpython

xxyzz:
I find only two lines of code set the node type to parser function:

https://github.com/tatuylonen/wikitextprocessor/blob/e5296c16f2d715e62121f23cb5057374da48cda3/wikitextprocessor/parser.py#L430

https://github.com/tatuylonen/wikitextprocessor/blob/e5296c16f2d715e62121f23cb5057374da48cda3/wikitextprocessor/parser.py#L919

wikitextprocessor/parser.py
kristian-clausal: I was wrong when I said the dev machine was using Python 3.8, that was my own local machine... The dev is on 3.10.

However, it is unlikely we will upgrade to 3.11 any time soon.

xxyzz:
Even though the above two lines change template kind to parser function, the extractor code still uses `isinstance(node, WikiNode)` and  `node.kind` to check the node type but not the class name, so I think this won't be a problem.

kristian-clausal:
At some point, we probably want to get rid of the node.kind stuff and only have classes, but yeah, in the meanwhile node.kind is fine.

wikitextprocessor/core.py
```diff
@@ -1692,37 +1692,46 @@ def get_page(
...
+                # Wikitext and Lua code
                 if title.startswith(ns_prefix):
                     template_name = title[len(ns_prefix) :]
-                    title = (
+                    lower_case_title = (
                         ns_prefix + template_name[0].upper() + template_name[1:]
                     )
                 else:
-                    title = ns_prefix + title[0].upper() + title[1:]
+                    lower_case_title = ns_prefix + title[0].upper() + title[1:]
```
kristian-clausal, line 1703: If you use .upper() here, then `lower_case_title` has the wrong name; it's an upper_case_title.

xxyzz: Thanks! I don't know why I named it `lower_case`...

xxyzz:
I'm trying to use the new class method and attribute in wiktextract and find the code at here also create `WikiNode`: https://github.com/tatuylonen/wiktextract/blob/205c4a2d88c27113f0117e0095f466605976af81/wiktextract/page.py#L80

PR #91, opened 2023-08-29
Update Scribunto code

xxyzz:
This pr changes the Scribunto folder to git submodule to add the new [`strict`](https://www.mediawiki.org/wiki/Extension:Scribunto/Lua_reference_manual#strict) module in Scribunto which is used by some Wikipedia Lua code. pip install will init git submodule automatically. But on server script, the git clone command need to add a new `--recurse-submodules` option.


kristian-clausal:
It looks good to me, and I will update the kaikki regeneration script when this gets merged; trouble is, the current run of kaikki failed (I've sent you the e-mail with the stack trace) in the new get-page-redirection code and I would like to get that working so that we don't have potential overlapping errors.

kristian-clausal (approved):

PR #92, opened 2023-08-29
Fix infinite recursion caused by "Template:cite-book"

xxyzz:
SQL query in `Wtp.get_page` returns page "Template:Cite-book" first and this page redirects to "Template:cite-book" again thus causes the infinite recusion.

### DIFF
--- a/tests/test_wikiprocess.py
+++ b/tests/test_wikiprocess.py
@@ -4155,6 +4155,15 @@ return export
         page = self.ctx.get_page("Template:En-nm", 10)
         self.assertTrue(page.need_pre_expand)
 
+    def test_get_page_resolve_redirect_infinite_recursion(self):
+        self.ctx.add_page("Template:cite-book", 10, body="cite-book")
+        self.ctx.add_page(
+            "Template:Cite-book", 10, redirect_to="Template:cite-book"
+        )
+        self.ctx.db_conn.commit()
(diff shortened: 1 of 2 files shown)


xxyzz:
Pages that redirect more than once can't be found now, I checked the db only these pages redirect twice:

```
sqlite> SELECT a.title, b.title, b.redirect_to FROM pages as a JOIN pages as b WHERE a.redirect_to = b.title AND b.redirect_to is not null;
Reconstruction:Proto-Turkic/yiŋč-ge|Reconstruction:Proto-Turkic/yiŋčge|Reconstruction:Proto-Turkic/yīnčge
Reconstruction:Proto-Tungusic/jalu(m)|Reconstruction:Proto-Tungusic/ǯalum|Reconstruction:Proto-Tungusic/ʒalum
Reconstruction:Proto-Tungusic/ǯube-|Reconstruction:Proto-Tungusic/ǯūr|Reconstruction:Proto-Tungusic/ʒūr
Template:Cite LGPSI|Template:quote-LGPSI|Template:RQ:LGPSI
```

but I can't find the last template page on Wiktionary, maybe it's not used and got deleted.

kristian-clausal:
With so few pages where this happens it's pretty obvious not how redirect is meant to be used and we can consider those double-redirects erroneous and ignore them.




🥳

PR #98, opened 2023-09-04
Return optional child node index number from `WikiNode.find_child()`

xxyzz:


kristian-clausal:
It would be helpful if you documented what all these changes are for; the whys and the whats at least. What is the optional child node index for? What is .find_child() used for? etc.

xxyzz:
I find these features are widely used in wiktextract code and add them to `WikiNode` would make writing extractor code easier(please see https://github.com/tatuylonen/wiktextract/pull/320). As the function names implied, `find_child()` is used for finding a certain type of child node. And child node index is the index of the children list, it could be used for divide child nodes. These functions are really simple, just a few lines and I think with the function names they could explain themselves.

kristian-clausal:
Please document what you do in any case. Do not assume anything is obvious. It is always easier to write code than to read it. Please add more comments. Please write longer explanations with whys and whats in commit messages. It is very frustrating having to piece things together when you could just have just written a few paragraphs of explanatory text. It is not difficult.

xxyzz:
I do add comments and example data in commit messages when needed but for these WikiNode methods I'm not sure what the comments should be. These are 5 or 6 lines of a for loop...

kristian-clausal:
The comment should be something like "This is for XYZ, it is not yet used anywhere but I am putting it here for when it is needed". Add the whys. If you're adding something that is not yet used, add the future use-case, otherwise it is a mystery what it is for.

xxyzz:
These are the best I can think of for now that could simplify extractor code without changing the parser data structure and breaking old code, not really for future usage, they could be used to replace some boilerplates right now. I'll add more comments. Maybe I should read some similar parser documents, I think the parser should have more select node methods.

PR #106, opened 2023-09-15
Pass namespace canonical name to Lua class `Namespace`

xxyzz:
Fixes Lua errors in some Arabic pages from French Wiktioanry. Example page: https://fr.wiktionary.org/wiki/ف ت ح

The error happends in https://fr.wiktionary.org/wiki/Module:arabe, line 1313: `mw.title.makeTitle( "template",racine):getContent()`. This Lua module uses the lower case canonical name, but our Lua code `mw_site.matchNamespaceName()` can't find it because the `canonicalName` is localized name("Modèle").

### DIFF
--- a/wikitextprocessor/lua/mw_site.lua
+++ b/wikitextprocessor/lua/mw_site.lua
@@ -21,7 +21,7 @@ Namespace.__index = Namespace
 function Namespace:new(obj)
    obj = obj or {}
    setmetatable(obj, self)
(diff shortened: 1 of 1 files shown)


kristian-clausal:
It looks good to me, but hopefully en.wiktionary modules don't rely on it being the other way around...

xxyzz:
It's won't affect English Wiktionary because English Wiktionary's namespace name is the also the canonical name.

PR #107, opened 2023-09-16
add support for python 3.8

empiriker:
In wikitextprocessor/core.py `from importlib.resources import files` is used. This is a Python 3.9 feature. It should either be replaced or Python 3.9 should be specified as a requirement as done via this commit.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,6 +43,7 @@ dependencies = [
     "lxml",
     "psutil",
     "requests",
+    "importlib_resources; python_version < '3.10'"
 ]
 
 [project.optional-dependencies]
--- a/wikitextprocessor/core.py
+++ b/wikitextprocessor/core.py
@@ -13,10 +13,13 @@ import sqlite3
 import sys
 import tempfile
 import urllib.parse
-from collections.abc import Sequence
+import sys
+if sys.version_info < (3, 9):
+    from typing import Sequence
+else:
+    from collections.abc import Sequence
 from dataclasses import dataclass
 from functools import lru_cache
-import sys
 if sys.version_info < (3, 10):
     from importlib_resources import files
 else:
--- a/wikitextprocessor/core.py
+++ b/wikitextprocessor/core.py
@@ -1034,20 +1034,19 @@ class Wtp:
         # refer to them
         while len(expand_stack) > 0:
             page = expand_stack.pop()
-            title_no_ns_perfix = page.title.removeprefix(
-                template_ns_local_name + ":"
-            )
-            if title_no_ns_perfix not in included_map:
+            prefix = template_ns_local_name + ":"
+            title_no_ns_prefix = page.title[len(prefix):] if page.title.startswith(prefix) else page.title
+            if title_no_ns_prefix not in included_map:
                 if self.lang_code == "zh":
-                    title_no_ns_perfix = (
-                        title_no_ns_perfix[0].lower() + title_no_ns_perfix[1:]
+                    title_no_ns_prefix = (
+                        title_no_ns_prefix[0].lower() + title_no_ns_prefix[1:]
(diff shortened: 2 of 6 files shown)


xxyzz:
`importlib.resources.files` actually requires 3.10(https://setuptools.pypa.io/en/latest/userguide/datafiles.html#accessing-data-files-at-runtime), older Python could use the "[importlib-resources](https://pypi.org/project/importlib-resources)" package. I should add the backport package back then...

empiriker:
I have added the backport package then.

pyproject.toml
```diff
@@ -43,6 +43,7 @@ dependencies = [
     "lxml",
     "psutil",
     "requests",
+    "importlib_resources"
```
xxyzz, line 46: This package should only be installed for older Python, like this: `importlib_resources; python_version < '3.10'`

wikitextprocessor/core.py
```diff
@@ -16,7 +16,7 @@
 from collections.abc import Sequence
 from dataclasses import dataclass
 from functools import lru_cache
-from importlib.resources import files
+from importlib_resources import files
```
xxyzz, line 19: You should check the python version(maybe using the `platform` library) and only use the backport package for older Python version.

pyproject.toml
empiriker: Thanks. I didn't know that pyproject.toml support conditional dependencies. Great feature!

wikitextprocessor/core.py
empiriker: Any reason why not to use ```if sys.version_info < (3, 10):```?

empiriker:
I ran the test suite successfully with python3.10.

I also experimented with 3.8. and the conditional import seemed to work. However, there are still more instances where the code actually assumes a higher python version.

For example:
`_nowiki_re: re.Pattern[str] = re.compile(...)` from common.py uses a 3.9 feature. 
[https://docs.python.org/3.11/library/re.html#re.Pattern](https://docs.python.org/3.11/library/re.html#re.Pattern)

empiriker:
I discovered several other incompatibilities with python 3.8 and resolved them.

I ran the tests with a python 3.8 and python 3.10 environment successfully.

xxyzz:
Honestly, I don't really care about 3.8 because I don't see any benefit using and support it and it's soon to be unmaintained. And I doubt anyone would have no other choice but have to use 3.8(no one complained so far and using docker is not that hard) and totally ignore all the performance improvements and new APIs of newer versions. But that's just my opinion.

empiriker:
Well, I tend to agree and had I known how much needed changing to add back support for 3.8, I might not have implemented the changes in this pull request here. But it was interesting nevertheless.

Regardless, either python 3.8 should be actually supported as is claimed by `pyproject.toml` or dropped as in #109 .

xxyzz:
I'm for dropping 3.8 because 3.8 will be dead next year, but I want to wait other maintainers' comments.

I'm not sure what our support policy is, if we support the last three major releases, we could also drop 3.9 because 3.12 will be released next month.

kristian-clausal:
Because we've been inadvertently using 3.9 features, I've merged the other pull request that drops support for 3.8. Sorry about the lost labor here. I will ask Tatu to formulate a more explicit policy regarding Python versions so that there are some guidelines when and when not to upgrade.

xxyzz:
If we still need to support 3.9 then the commit https://github.com/tatuylonen/wikitextprocessor/pull/107/commits/0bcb4f1e5e66082dcb3ac634a1b3d0d035b58d88 could be used.

kristian-clausal:
Sorry, missed that. I don't think I could partially merge just that one commit (or at least I'm not brave enough to try), so I'll just open this again.

PR #109, opened 2023-09-17
drop support for python 3.8 and require 3.9

empiriker:
Currently, the code is not compatible with python 3.8, since 3.9 features are frequently used. This commit drops support for python 3.8 and requires python 3.9 via pyproject.toml.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,6 @@ classifiers = [
     "Operating System :: POSIX :: Linux",
     "Programming Language :: Python",
     "Programming Language :: Python :: 3 :: Only",
-    "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",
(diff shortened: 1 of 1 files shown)


empiriker:
Why I care about this: I was recently helping a colleague who wanted to run wiktextract using python3.8 and ran into problems.

If 3.8 was required as in this pull request, he would have gotten the warning:
`ERROR: Package 'wikitextprocessor' requires a different Python: 3.8.18 not in '>=3.9'`
when installing the wiktextract and it's dependency wikitextprocessor.

See also #107.

kristian-clausal:
If we've been inadvertently using features from 3.9, I guess we really should stop 3.8 support. I will ask Tatu to formulate a policy about Python version upgrades, so that we have some guidelines to follow in this, when needed. Please don't start using Python 3.14 features just so that we are forced to upgrade, lol.

PR #112, opened 2023-09-21
Support all Page class attibutes in overwrite JSON file

xxyzz:
Only page body was used before, now the overwitre JSON file could adds more page data in a dictionary. For example:

```json
{
  "Template:title": {
    "body": "page_body",
    "namespace_id": 10,
    "redirect_to": "redirect_page_title",
    "need_pre_expand": true,
    "model": "wikitext"
  }
}
```

All fields are optional.


kristian-clausal:
Looks good to me. :+1:

PR #114, opened 2023-09-28
Remove some title checking rules in `mw_title.makeTitle()`

xxyzz:
MediaWiki code doesn't have these checks, I only removed 3 lines to fix Lua errors in French Wiktionary.


kristian-clausal:
There are probably reasons why Tatu implemented this this way, but then again sometimes the original is slightly wrong. Let's keep an eye out if this breaks something big on the en.wiktionary side.

xxyzz:
Because the code was moved from wiktextract it's kind hard to find which commit added these rules, git blame would only find the first commit moving the entire file.

kristian-clausal:
That kind of detective work is the worst.

PR #120, opened 2023-10-12
Fix regex characters range "A-Z"

xxyzz:
This error was found by GitHub Code Scan, here is the screenshot from my forked repo:

<details>
<summary>screenshot</summary>

image
</details>

### DIFF
--- a/src/wikitextprocessor/core.py
+++ b/src/wikitextprocessor/core.py
@@ -534,7 +534,7 @@ class Wtp:
                     # re.X = ignore whitespace and comments, re.I = ignore case
                     r"""(?xi)\|(
                             (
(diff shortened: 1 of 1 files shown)


kristian-clausal:
Oh, that was a nasty little one. Good find, hopefully it didn't cause any problems.

xxyzz:
GitHub Code Scanning also finds some regex backtracking problems in the "src/wiktextract/clean.py" file, but these regex patterns are more complicated, I'll check that file tomorrow.

kristian-clausal:
Backtracking problems are bad, but it is possible that they have never triggered except in exceptional circumstances. If you can figure them out, that would be great.

PR #122, opened 2023-10-13
Add an option to skip analyzing templates

xxyzz:
Pre-expand templates is not very useful for the French Wiktionary. For example, translation start template "trad-début" doesn't need to be expanded and it's sense parameter would be lost if expanded. And it also moves the translation list nodes inside some HTML tags.


kristian-clausal:
Looks good to me! :+1:

PR #124, opened 2023-10-17
Match latex math formula in template parameters

xxyzz:
Previous template regex pattern can't match templates with parameter that has a "{" right after "}". Also remove the table pattern to keep the pattern simple, tables are never used in template parameters.

Resolves https://github.com/tatuylonen/wiktextract/issues/365.

Pray this change won't break anything...


kristian-clausal:
This looks fine, but regex is as regex does so we'll have to see if it explodes later. :+1:

PR #125, opened 2023-10-19
Restore regex patterns match unpaired "{" and "}" in template body

xxyzz:
Commit f815e426ddf4a8818182662af84cf11d72e4cc1a removes these two patterns causes the code can't match the `#switch` parser function in template "str index-lite/logic" and generates a very long expanded text, then triggers regex exponential backtracking at file clean.py line 1444.


kristian-clausal:
This looks good to me, regex can be such a pain to work with. You never know what is there on purpose and what's just cruft left over from older versions of a regex string, because they are so annoying to read and edit after-the-fact. Good job.

PR #127, opened 2023-10-23
Change `NodeKind` to `enum.Flags`

xxyzz:
to support selecting multiple node types

### DIFF
--- a/src/wikitextprocessor/parser.py
+++ b/src/wikitextprocessor/parser.py
@@ -105,7 +105,7 @@ MAGIC_WORDS: Set[str] = set(
 
 
 @enum.unique
-class NodeKind(enum.Enum):
+class NodeKind(enum.Flag):
     """Node types in the parse tree."""
 
     # Root node of the tree.  This represents the parsed document.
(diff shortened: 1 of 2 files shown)


kristian-clausal:
Seeing "in target_kinds" like this looks like target_kinds should be a list. Needs some documentation (in class NodeKind and places where the "flagginess" of NodeKind is used) to make it explicit you have to use the enum.Flag `|` or-thing, not a list. I had no idea enum.Flag was even a thing (although it's obvious in hindsight), and would have had a bit of trouble if I made assumptions here... Just a couple of comments indicating NodeKind is now a bitmask thing should be fine.

xxyzz:
The argument types are still `NodeKind` not a list, mypy or pydantic would find error when a list is passed.

I think it's quite natural to combine enum types, change to `Flag` class doesn't affect the previous code, a single enum type could still be passed and it also doesn't affect the code compare to a single type. This is a proper use of the enum type. And I don't think the flaggable type could be misused.

I'm adding a change to the French extractor's linkage code, which need to find template nodes and list nodes under the level node, for example:

```
==== {{S|synonymes}} ====
{{(|Celle qui est à l’origine de quelque chose|1}}
* [[artisane]]
```

It'd be very convenient if `WikiNode.find_child()` supports selecting multiple node kinds.

xxyzz:
I add a comment of the new flaggable argument type in this commit: https://github.com/tatuylonen/wikitextprocessor/pull/128/commits/0d42fc229036acfd9c5328f0a265e71dc3741a3d

PR #132, opened 2023-10-31
Fix two catastrophic backtracking cases when matching wikitext table

xxyzz:
This bug was found in https://github.com/tatuylonen/wiktextract/issues/389, but this pr only fixes two cases. `Wtp._analyze_template()` sill has more backtracking bugs while processing the Russian Wikipedia templates. The code now stuck at page https://ru.wikipedia.org/wiki/Шаблон:Результаты_зимних_Олимпийских_игр_2006, this template seems to use the `{{!}}` template at the end of the table.


kristian-clausal (approved):
Excellent catch, I wouldn't have realized this.

PR #133, opened 2023-10-31
Update Spanish namespace data

empiriker:
I started to look at the Spanish Wiktionary and noted that the namespaces are not named correctly. Especially the Appendix namespace which seems to hold the best source of language data.

I have added a format commit so that at least the diff of the last commit allows to see which namespace data has changed.


kristian-clausal:
Thank you, one step forward!

xxyzz:
Thanks! Most of the namespace JSON files are not created from the "tools/get_namespaces.py" script and need to be reviewed when implementing new extractor.

PR #134, opened 2023-11-01
Get language names from ICU

xxyzz:
`mw.language.fetchLanguageName` and `mw.language.fetchLanguageNames` use the language data extracted from CLDR(https://cldr.unicode.org)

I can't find the ICU API to get language code from language name, but I guess we could use `getAvailable()` to build a dictionary. Then we could use the icu package to replace the language JSON files in the wiktextract package.

pyicu's doc says the `libicu-dev` package is required on Debian: https://gitlab.pyicu.org/main/pyicu#installing-pyicu

### DIFF
--- a/src/wikitextprocessor/luaexec.py
+++ b/src/wikitextprocessor/luaexec.py
@@ -242,24 +242,29 @@ def get_page_content(ctx: "Wtp", title: str) -> Optional[str]:
     return ctx.read_by_title(title.strip())
 
 
-def fetch_language_name(ctx: "Wtp", code: str) -> str:
-    """This function is called from Lua code as part of the mw.language
-    implementation.  This maps a language code to its name."""
-    ret = ctx.LANGUAGES_BY_CODE.get(code)
-    if ret:
-        return ret[0]
-    return ""
+def fetch_language_name(code: str, inLanguage: Optional[str]) -> str:
+    """
+    This function is called from Lua code as part of the mw.language
+    implementation.  This maps a language code to its name.
+    https://www.mediawiki.org/wiki/Extension:Scribunto/Lua_reference_manual#mw.language.fetchLanguageName
+    """
+    from mediawiki_langcodes import code_to_name
+
+    return code_to_name(code, inLanguage or "")
 
 
-def fetch_language_names(ctx: "Wtp", include: str) -> "_LuaTable":
+def fetch_language_names(
+    ctx: "Wtp", inLanguage: Optional[str], include: Optional[str]
+) -> "_LuaTable":
     """This function is called from Lua code as part of the mw.language
     implementation.  This returns a list of known language names."""
-    include = str(include)
-    if include == "all":
-        ret = ctx.LANGUAGES_BY_CODE
-    else:
-        ret = {"en": ctx.LANGUAGES_BY_CODE["en"]}
-    return ctx.lua.table_from(ret)  # type: ignore[union-attr]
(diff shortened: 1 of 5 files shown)


src/wikitextprocessor/luaexec.py
```diff
@@ -242,24 +242,38 @@ def get_page_content(ctx: "Wtp", title: str) -> Optional[str]:
...
+) -> str:
+    """
+    This function is called from Lua code as part of the mw.language
+    implementation.  This maps a language code to its name.
+    https://www.mediawiki.org/wiki/Extension:Scribunto/Lua_reference_manual#mw.language.fetchLanguageName
+
+    pyicu: https://gitlab.pyicu.org/main/pyicu
+    icu: https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/index.html
+    """
+    from icu import Locale
```
kristian-clausal, line 256: Googling it, there doesn't seem to be much of a difference in resource-use or speed whether imports happen at the start of a module or inside a function (because the import is cached anyhow), but we don't do much of this kind of importing inside function elsewhere in the code so it feels better to move this to the start of the file with the others, as `import icu` so that it doesn't become too unclear that `icu.Locale` isn't an import when the import isn't inside the function.

xxyzz:
`fetchLanguageNames` returns languages MediaWiki "knows" if `include` is not "all", I find the MediaWiki code: https://github.com/wikimedia/mediawiki/blob/af780552ab093d47047f561a0c82051b883e2116/includes/language/LanguageNameUtils.php#L257-L261

I guess that means it returns languages in this file: https://github.com/wikimedia/mediawiki/blob/master/includes/languages/data/Names.php#L41

but I just return all the language data.

kristian-clausal (approved):
Moving over from our own kludgy data files to a maintained one-purpose library that is a shared source with Wiktionary is good, and this seems correct.

src/wikitextprocessor/luaexec.py
xxyzz: Python doesn't have lazy import, import dependencies when using them would decrease startup time. But I add the import inside function mainly because it's external package and it's only used in these two functions. I think other dependencies like `lxml`, `requests` and `pyutil` are also imported in this way.

xxyzz:
@kristian-clausal the previous code assumes `mw.language.fetchLanguageName` only takes one argument, you might need to filter some false alarms from these two functions in your testing.

kristian-clausal:
> @kristian-clausal the previous code assumes `mw.language.fetchLanguageName` only takes one argument, you might need to filter some false alarms from these two functions in your testing.

Luckily, I don't think fetchLanguageName has been "switched" around to get wrong arguments, but yeah, need to remove the debug statements there.

xxyzz:
I find the language data use in the Lua API that are not included in ICU: https://github.com/wikimedia/mediawiki-extensions-cldr/tree/master/LocalNames, and https://github.com/wikimedia/mediawiki/blob/master/includes/languages/data/Names.php

Extract the data would need to run some php code.

xxyzz:
I create a package for converting mediawiki language names and codes: https://github.com/xxyzz/mediawiki_langcodes

There are 157 language names have more than one language codes, I haven't handle these cases.

xxyzz:
For language names in English there are some rare dialect names have multiple language codes, most are added from Wiktionary language titles not from CLDR, and `name_to_code` chooses the shortest code.

PR #137, opened 2023-11-09
Match HTML attibutes that have white spaces around `=`

xxyzz:


kristian-clausal:
Good catch, HTML syntax details can be so annoying because people 99.999% do it a specific way, but then there's cases like these where it's actually more forgiving than you'd think.

PR #142, opened 2023-11-16
Clear Lua environment for each `{{#invoke:}}`

xxyzz:
Scribunto does this: https://www.mediawiki.org/wiki/Extension:Scribunto#Lua_environment

### DIFF
--- a/src/wikitextprocessor/luaexec.py
+++ b/src/wikitextprocessor/luaexec.py
@@ -444,24 +444,20 @@ def call_lua_sandbox(
         )
         return "{{" + invoke_args[0] + ":" + "|".join(invoke_args[1:]) + "}}"
 
-    # Initialize the Lua sandbox if not already initialized
-    if ctx.lua_depth == 0:
-        if ctx.lua is None:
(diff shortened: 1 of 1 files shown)


kristian-clausal:
Is there a possibility that there can be recursive invoke-calls? Isn't that what ctx.lua_depth is for? If so, wouldn't this basically wipe the outer scope, so the previous environment is reset when the context returns from inner `invoke`s?

xxyzz:
Maybe our implementation is incorrect? The Scribunto doc says:

>Each {{#invoke:}} call runs in a separate environment. Variables defined in one {{#invoke:}} will not be available from another. This restriction was necessary to maintain flexibility in the wikitext parser implementation.

And I think this is their code: https://github.com/wikimedia/mediawiki-extensions-Scribunto/blob/8d69dc173e33ae936ff4401d41ee5e6a1fd1ba67/includes/Engines/LuaCommon/lualib/mw.lua#L458-L467

This commit might need to be reverted if it breaks some pages.

PR #143, opened 2023-11-16
Clone Lua env for each `#invoke`

xxyzz:
- revert #142
- clone Lua env for each `#invoke`

### DIFF
--- a/src/wikitextprocessor/lua/_sandbox_phase1.lua
+++ b/src/wikitextprocessor/lua/_sandbox_phase1.lua
@@ -12,10 +12,41 @@ local loader_cache = {}
 local loaddata_cache = {}
 local _orig_package = package
 
+-- Copied from https://github.com/wikimedia/mediawiki-extensions-Scribunto/blob/8d69dc173e33ae936ff4401d41ee5e6a1fd1ba67/includes/Engines/LuaCommon/lualib/mwInit.lua#L38-L66
+--- Do a "deep copy" of a table or other value.
+function mw_clone( val )
+    local tableRefs = {}
+    local function recursiveClone( val )
+        if type( val ) == 'table' then
+            -- Encode circular references correctly
+            if tableRefs[val] ~= nil then
+                return tableRefs[val]
+            end
+
+            local retVal
+            retVal = {}
+            tableRefs[val] = retVal
+
+            -- Copy metatable
+            if getmetatable( val ) then
+                setmetatable( retVal, recursiveClone( getmetatable( val ) ) )
+            end
+
+            for key, elt in pairs( val ) do
+                retVal[key] = recursiveClone( elt )
+            end
+            return retVal
+        else
+            return val
+        end
+    end
+    return recursiveClone( val )
+end
+
 -- This function loads new a new module, whether built-in or defined in the
 -- data file, and returns its initialization function.  This caches the
 -- initialization function.
-function new_loader(modname)
+function new_loader(modname, clone_env)
    -- this module wreaks havoc on the global namespace
    if modname == "Module:No globals" then
       return nil, "Module banned"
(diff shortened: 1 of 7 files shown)


kristian-clausal:
This seems like the way to go, afaict; it should at the very least be a bit more correct than what we're currently doing. My only concern is the cost of the cloning, but 'needs must' (that's a real English phrase, rare and super-wonky).

kristian-clausal (approved):

kristian-clausal:
Sorry, I accidentally pressed the button that stops this from being a draft! shouldn't break anything, hopefully...

xxyzz:
But this change causes Lua errors in French Wiktionary page [responsable des services généraux](https://fr.wiktionary.org/wiki/responsable_des_services_g%C3%A9n%C3%A9raux): 
```régulateur de vitesse et d’espacement/Français/Étymologie: ERROR: LUA error in #invoke('section', 'section') parent ('Modèle:S', {1: 'nom', 2: 'fr'}) at ['régulateur de vitesse et d’espacement', 'S', '#invoke']
[string "Module:locution"]:51: attempt to index global 'typ' (a nil value)
```

xxyzz:
Maybe it's because cached Lua module has different env?

xxyzz:
Still incorrect... The following loaded Lua modules are not using the cloned env.

kristian-clausal:
The problem is difficult, because the differences between our implementation and the original, and just in general. In the meanwhile, if you're working on this in this PR/branch we should probably revert the commit from this morning for the Kaikki runs over the weekend. The run we started this morning is still going on, but grepping the logs there's over a million instances of `LUA error` in the log, so I'm going to interrupt that and revert this morning's patch.

xxyzz:
Maybe there is a reason that Scribunto clones Lua env in `#invoke`. I found one of the changes in Lua 5.2 is the environment: https://www.lua.org/manual/5.2/manual.html#8.1, perhaps we don't have the memory bug before because we're not using 5.1 back then. I don't know whether clone env would fix the memory bugs but I think it's worth trying.

src/wikitextprocessor/lua/_sandbox_phase1.lua
```diff
@@ -12,20 +14,55 @@ local loader_cache = {}
...
+    end
+    return recursiveClone( val )
+end
+
 -- This function loads new a new module, whether built-in or defined in the
 -- data file, and returns its initialization function.  This caches the
 -- initialization function.
-function new_loader(modname)
-   -- this module wreaks havoc on the global namespace
-   if modname == "Module:No globals" then
```
xxyzz, line 20: This Lua module is removed from Wiktionary and this code was added because Lua environments are not isolated(https://github.com/tatuylonen/wikitextprocessor/commit/faa4bfce87cd7cfc37f0a2b5d39460892d0c48ce).

xxyzz:
I guess `test_mw_title59` failed because the result of `mw.title.getCurrentTitle` is cached in `package.loaded`? If I comment the cache code in `new_require()`: https://github.com/tatuylonen/wikitextprocessor/blob/0112b95943293a592e11d915af23e9e7369d1a4a/src/wikitextprocessor/lua/_sandbox_phase1.lua#L74-L77

xxyzz:
Extract the English db file takes about 3 hours, I think this pr doesn't slow down the performance. I searched the log and didn't find `LAMBDA` warnings and exceptions, the Lua errors seem like our usual errors.

kristian-clausal:
This seems excellent. If the performance doesn't dip (too much), then changing over to this from the unitary environments seems like the correct thing to do, even if it doesn't help with the function scrambling/memory scrambling stuff. If nothing is broken in the output, then I don't see why we couldn't merge this after you're finished with it and see what happens with kaikki. The worst case scenario is that the memory issues persist, and get worse, but that seems unlikely, or that this introduces some other sort of big bug that just didn't show, but that also seems unlikely.

xxyzz:
The last two extract jobs on kaikki.org don't have that random Lua memory errors, was the sever rebooted recently?

This pr has some other changes might relate to the memory bug, like remove the `_mw_frame` global variable, and replace some Python `lambda` functions with `functools.partial()`

kristian-clausal:
The memory errors on kaikki.org are random, and there can be gaps between them manifesting. That's why it's been so impossible to figure out.

PR #144, opened 2023-11-21
Impliment `#statements` parser function

xxyzz:


kristian-clausal:
Thank you for implementing all of this. That linter is really strict if it really fails on a typo in a comment...

xxyzz:
I don't think that's strict, typo in comments is still typo after all...

PR #152, opened 2023-12-01
Add test of auto newline added before parser function

xxyzz:


kristian-clausal:
Of course, the test applies to parserfns! I tried to figure out how to do it with `multitrans`, and decided to leave it for later... Thanks!

xxyzz:
What's wrong with the `multitrans` template? I checked the page `dictionary` which uses `multitrans` and the translation lists are parsed correctly.

PR #153, opened 2023-12-04
Clear the timeout hook after a Lua model is finished execution

xxyzz:
I got a strange Lua timeout error from the 20231201 Chinese Wiktionary dump file on page "还了得", a simple page. The Lua error is at the `_lua_reset_env()`, before running the Lua module code.


xxyzz:
I think it only happens once, I didn't rerun the command. But the timer should be cleared otherwise it's still ticking after the Lua model code is finished and could stop other Lua code.

PR #157, opened 2023-12-08
Pass namespace id to `Wtp.get_page()` as much as possible

xxyzz:
I suspect this function could return wrong result from cache when namespace id is not passed

### DIFF
--- a/src/wikitextprocessor/core.py
+++ b/src/wikitextprocessor/core.py
@@ -1445,7 +1445,9 @@ class Wtp:
                     name = tname
 
                     # Check for undefined templates
-                    if not self.template_exists(name):
+                    if not self.page_exists(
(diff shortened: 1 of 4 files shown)


src/wikitextprocessor/core.py
```diff
@@ -1814,16 +1810,8 @@ def check_template_need_expand(
 
         return page.need_pre_expand
 
-    def read_by_title(
-        self, title: str, namespace_id: Optional[int] = None
-    ) -> Optional[str]:
-        """Reads the contents of the page.  Returns None if the page does
-        not exist."""
-        page = self.get_page_resolve_redirect(title, namespace_id)
-        return page.body if page is not None else None
```
xxyzz, line 1823: This old method get the name because it reads a huge cache file before we migrate to sqlite db. The new  `get_page_body()` now requires namespace id.

PR #160, opened 2023-12-11
Remove code added from commit 6d2d466b2a640961c68cb93db6330f40cdf8a588

xxyzz:
I'm not sure what error that commit was trying to fix, the current code has no problem process the translation section of the "quadratic" page and the "trans-top" template's body is `{{#invoke:translations|top}}`. Even if a template parameter is passed to `#invoke` and it contains `=` the code is still correct and don't need to replace `=`.

Remove this code fixes the Lua error in page "can" because `=` is replaced to `&#61;` and causes a error in module "translations/multi".

### DIFF
--- a/src/wikitextprocessor/core.py
+++ b/src/wikitextprocessor/core.py
@@ -1284,25 +1284,6 @@ class Wtp:
                             k = re.sub(r"\s+", " ", k).strip()
                         v = argmap.get(k, None)
                         if v is not None:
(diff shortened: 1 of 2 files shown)


kristian-clausal:
If later changes have fixed the original issue, then this is obviously correct to do. There's probably many other kludges like this still left, even if they are not relevant anymore.

PR #163, opened 2023-12-18
Save interwiki maps data to SQLite db file

xxyzz:
Call the `siteinfo` MediaWiki API only once and share the data across worker processes.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,5 +5,5 @@ updates:
     schedule:
       interval: "weekly"
     ignore:
-      - dependency-name: "typos"  # ignore patch updates
+      - dependency-name: "crate-ci/typos"  # ignore patch updates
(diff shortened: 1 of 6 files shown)


kristian-clausal:
'Caching' stuff we take from the web is sensible. Could this also be done with wikidata? Although with that we have a problem with timing; currently we're pulling WikiData stuff when processing pages. I guess it could be possible to pre-process pages during the dump to find wikidata links and 'cache' them there, that way we also "freeze" the data which might be updated later and get out-of-sync with the dump itself (which doesn't change).

xxyzz:
I'll create a new pr to save wikidata to sqlite db. But we don't have any issues with wikidata for now, unlike the siteinfo API, the previous code calls this API many times when processing the Chinese Wiktionary dump file.

Wikidata could be used in Lua code, even search for the `statement` parser function might not be easy.

PR #164, opened 2023-12-18
Save wikidata API results to SQLite db file

xxyzz:


kristian-clausal:
Thank you this, this look great!

xxyzz:
I haven't run the code on a dump file, will do that tomorrow, in case there are bugs not triggered by tests.

xxyzz:
The function `get_interwiki_map()` causes the TypeError happens to be mocked in test so I didn't catch the error in my last pr...

PR #165, opened 2023-12-18
Type checking 3.9

kristian-clausal:
Updating the syntax to Python 3.9 type hinting, also attempted to take care of a few 3.9-specific issues (like function signatures that have been changed since 3.9, but still have a 3.9-friendly 'form' that can be used).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -78,6 +78,7 @@ wikitextprocessor = [
 
 [tool.mypy]
 mypy_path = "typestubs"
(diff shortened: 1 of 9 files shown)


xxyzz:
Jeez... If I find myself struggling to get type checker happy, I'll tell it to ignore the code.

Maybe we could only support the latest Python version like the home assistant project since we have container image now. But I guess you'd disagree. Write backward compatible Python code is not fun...

PR #166, opened 2023-12-20
Format code and enable `ruff` checks in lint action

xxyzz:
- I didn't know ruff could format code before, remove `black`.
- Use `Any`  in `TemplateNode.template_parameters()` to simplified the code, one of the `if` check is incorrect: line 546, named parameter name is always `str`.

Type check for the `requests` and `lxml` library almost useless...


src/wikitextprocessor/parser.py
```diff
@@ -541,22 +536,19 @@ def template_parameters(
                                 )
                             continue
 
-                if (is_named and 
-                        (isinstance(parameter_name, int) 
-                        or len(parameter_name)) > 0
-                ):
+                if is_named and len(parameter_name) > 0:  # type: ignore
```
kristian-clausal, line 546: I missed the `is_named = False` after the line where parameter_name is cast into int, so this is correct.

src/wikitextprocessor/parser.py
```diff
@@ -541,22 +536,19 @@ def template_parameters(
...
-                final_parameters[p_name] = p_value[0]
+                parameters[p_name] = p_value[0]
             else:
-                final_parameters[p_name] = p_value
+                parameters[p_name] = p_value
 
-        self._template_parameters = final_parameters 
-        return final_parameters
+        self._template_parameters = parameters
+        return parameters
```
kristian-clausal, line 558: We are returning defaultdict() here, which is why I changed around this bit of code in the first place, not just type-checking. If some other pieces of code receives a defaultdict() that returns empty values when it should throw an exception instead, that's going to be really annoying to debug. Please read my commit messages and don't just revert everything!

kristian-clausal (changes requested):

src/wikitextprocessor/parser.py
xxyzz: Sorry I miss the commit message, I convert it to normal dictionary now.

You see this function already makes you a little confused, it shouldn't be more complicated.

kristian-clausal (approved):

xxyzz:
I forget to remove the `else` branch...

kristian-clausal:
Thanks for the changes, this looks good. In retrospect, I should have put a comment in the code explaining why I was doing the defaultdict() -> dict() thing, and I will try to be better at commenting stuff in the future.

PR #169, opened 2023-12-21
Clean up `ns` parser function code

xxyzz:
`Wtp.namespaces` is only used in `ns_fn`, create another dictionary to save namespace data seems unnecessary.


kristian-clausal:
Please write a commit message that explains what is going on.

PR #171, opened 2023-12-22
Allow frozenset type passed to `Wtp.expand()` and `Wtp.parse()`

xxyzz:
Chinese extractor uses `frozenset` and failed at these lines. I enabled optimizations and `assert` are removed in this mode so I didn't find the error before. This is a bit annoying. All pages failed...

### DIFF
--- a/src/wikitextprocessor/core.py
+++ b/src/wikitextprocessor/core.py
@@ -13,7 +13,7 @@ import sys
 import tempfile
 import urllib.parse
 from collections import defaultdict, deque
-from collections.abc import Sequence
+from collections.abc import Sequence, Set
 from dataclasses import dataclass
 from functools import lru_cache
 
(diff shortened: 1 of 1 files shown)


kristian-clausal:
I tried to find some collections.abc abstract type that encompasses both set and frozenset, but there's only Set (frozenset) and MutableSet (set); all the other abstract types are probably too broad. This was in 3.9.

xxyzz:
`Set` can be used for mutable set. mypy check passed.

xxyzz:
The document doesn't say this but the abc library code shows `MutableSet` is a subclass of `Set`, so `Set` can be used for both `set` and `frozenset`: https://github.com/python/cpython/blob/c31943af16f885c8cf5d5a690c25c366afdb2862/Lib/_collections_abc.py#L702

PR #174, opened 2024-01-02
Transclude any page as template

xxyzz:
Any page can be used as template:
https://www.mediawiki.org/wiki/Help:Templates#Usage

This feature is used in French Wiktionary's Conjugaison page.


kristian-clausal:
Good catch.

PR #176, opened 2024-01-04
Allow apostrophe in template parameter name

xxyzz:
Find this in French Wiktionary's Conjugaison page.

Hopefully this won't break anything...


kristian-clausal:
This seems like the correct choice, although I wonder if it could cause trouble with italics/bold parsing. I couldn't quickly find anything about parameter name allowed characters, but I'll keep looking.

PR #177, opened 2024-01-05
Fix 'WikiNode' object has no attribute 'strip' error

xxyzz:
Can't expand nodes at here, and this error happens on a misuse of link as template name. Not sure if there are real usage of template for a template name, probably not very common.


kristian-clausal:
Good catch :+1:

PR #178, opened 2024-01-08
Don't match HTML end tag as start tag in arguments

xxyzz:
The previous regex pattern matches `<nowiki />` as a start tag therefore includes the second "if" argument into the first argument.

### DIFF
--- a/src/wikitextprocessor/core.py
+++ b/src/wikitextprocessor/core.py
@@ -529,11 +529,11 @@ class Wtp:
                     # re.X = ignore whitespace and comments, re.I = ignore case
                     r"""(?xi)\|(
                             (
-                                <([-a-zA-Z0-9]+)\b[^>]*>  # html tag
-                                    [^][{}]*?             # element contents
-                                                          # (including `|`'s)
-                                    </\3\s*>              # end tag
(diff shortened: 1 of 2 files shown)


kristian-clausal:
At first glance this fix looks correct to me, except I wonder if you can have `/` inside tags. The tokenization of `<tag />` is that you can't have whitespace between `<` and `tag` or between `/` and `>`; `/>` is one whole token, which might indicate that you can have `/` otherwise.

https://html.spec.whatwg.org/multipage/syntax.html#syntax-attributes

According to this, you *can* have an attribute value without quotes that contains `/`! I didn't know that, that's scary. So `<htmltag attribute=/ ></html> is completely valid. Wild! Is it possible to use ` <([-a-zA-Z0-9]+)\b[^>]*[^/]?>` without it being weird? No, that doesn't work, because it requires at least one character there that is not part of the tag name...

kristian-clausal:
Yeah, I think this requires lookbehind, `<([-a-zA-Z0-9]+)\b[^>]*(?<!/)>`.

xxyzz:
Thanks! More common case is "/" inside the attribute value like a link and the pattern can't match that tag, I didn't think of this.

PR #181, opened 2024-01-11
Only match table row token at the beginning of a line

xxyzz:
table cell text might also starts with `-`

### DIFF
--- a/src/wikitextprocessor/parser.py
+++ b/src/wikitextprocessor/parser.py
@@ -1977,7 +1977,7 @@ header_re = re.compile(r"(?m)^(={1,6})\s*(([^=]|=[^=])+?)\s*(={1,6})\s*$")
 
 # Regular expression for matching a token in WikiMedia text.  This is used for
 # tokenizing the input.
-token_re = re.compile(
+TOKEN_RE = re.compile(
     r"'''''|"
     r"'''|"
     r"''|"
@@ -1986,8 +1986,8 @@ token_re = re.compile(
     r"\]|"
     r"\|\}|"
     r"\{\||"
-    r"\|\+|"
-    r"\|-|"
+    r"^\s*\|\+|"  # table caption
+    r"^\s*\|-|"  # table row
     r"!!|"
     r"\s*https?://[a-zA-Z0-9.]+(/[^][{}<>|\s]*)?|"
     r"^[ \t]*!|"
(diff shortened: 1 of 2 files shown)


kristian-clausal:
Good catch!

kristian-clausal:
Whitespace is allowed before the `|-`, but `<nowiki />|-` does not create a new row.

kristian-clausal:
`|` at the beginning of a row also allows whitespace before it. All of the ones that are meant to start a line allow whitespace before them...

kristian-clausal:
Python regex look-behind patterns (`(?<=...)` require fixed width patterns. And we can't just use .strip(), because we capture whitespace elsewhere, and capture groups with `|` regex separators are not useful (because group(1) is always the first capture group *in the pattern*, not the match, and the .groups() is a list of all the possible capture groups in the pattern in the order which they appear in the pattern, with 'None' if they don't match).

kristian-clausal:
I think I found a bug in how `|` is being handled, but it's tricky (probably because this looks like my own spaghetti-code...). `|` as a cell separator at the beginning of a line allows whitespace, but it seems we don't allow whitespace in vbar_fn that handles `|` and `||`... This might require a bigger job.

xxyzz:
I updated the pattern to allow white spaces before  `|+` and `|-`. I didn't change other patterns because there are two empty table tests would fail.

kristian-clausal:
We need to change the tests, because if you try out tables in https://www.mediawiki.org/w/index.php?title=Project:Sandbox&action=submit and put whitespace before the tokens, they don't break things. This might actually be best handled in the token functions, not the regex; the regex can collect the tokens and the token functions validate them, like vbar_fn does. I'm taking a look at things, but it's a bit spaghettuous...

xxyzz:
I find two failed tests: https://github.com/tatuylonen/wikitextprocessor/blob/0cb88feb77be2e03ba0eac4aca1ea1cb4cba20f8/tests/test_parser.py#L1541-L1542 and https://github.com/tatuylonen/wikitextprocessor/blob/0cb88feb77be2e03ba0eac4aca1ea1cb4cba20f8/tests/test_node_expand.py#L145-L146

"test_table_empty" is correct, not sure the other one. Should I close this pr?

kristian-clausal:
Yeah, I'll take a look at this. It's going to be so annoying to figure out... Or so simple that I'm going palm my face. Either one. Thanks for pointing this out!

kristian-clausal:
We should reopen this and take the test from it, I've tried to fix the issue with these tokens in #182. There, the whitespace is handled in the token functions instead of in the regex.

PR #182, opened 2024-01-11
Accept whitespace before certain table tokens

kristian-clausal:
Previously, tokens such as `{|`, `|+`, `|-` and `|}` were allowed to function as tokens for parsing tables even when they weren't at the beginning of a line, or following whitespace at the beginning of a line. This is incorrect.

`|` and `!` (separating headers and cells) when singular need to be at the start of the line or after start of line whitespace, but not `||` and `!!` which are used within a line. `|` is a bit more complicated because the `|` token is also used an attribute separator in tables!

Had to change a few mistaken tables. `{| |}` results in broken tables because the `|}` is not parsed without a newline, and our parser will consume the rest of the section or document. Wikimedia's parser does something weird, and in this case the empty table above (without a proper end token, so it's parsed wonkily) moves forward in text until it encounters the end of the document or another table and then puts an empty table. So if you have a string after the text above, the string is *before* the empty HTML-table. It's weird, and we will at least for now keep this edgecase as it is.

### DIFF
--- a/src/wikitextprocessor/parser.py
+++ b/src/wikitextprocessor/parser.py
@@ -1259,13 +1259,27 @@ def colon_fn(ctx: "Wtp", token: str) -> None:
     node.children = []
 
 
+def mistokenized_start_fn(ctx: "Wtp", token: str) -> None:
+    """Handler for table start token or text + double pipe toke."""
+    if ctx.pre_parse:
+        return text_fn(ctx, token)
+
+    if not (ctx.beginning_of_line or ctx.wsp_beginning_of_line):
+        text_fn(ctx, "{")
+        return double_vbar_fn(ctx, "||")
+
+    table_start_fn(ctx, "{|")
+    return vbar_fn(ctx, "|")
+
+
 def table_start_fn(ctx: "Wtp", token: str) -> None:
     """Handler for table start token "{|"."""
     if ctx.pre_parse:
         return text_fn(ctx, token)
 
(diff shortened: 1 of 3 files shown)


xxyzz:
If  "|-" cell token is parsed as plain text, then the test in #181 would still fail?

kristian-clausal:
> If "|-" cell token is parsed as plain text, then the test in #181 would still fail?

It should pass, because `|-` is at the beginning (or after whitespace at the beginning) of a line.

xxyzz:
Do you mean the test will pass because the regex pattern is changed? That test would fail if token regex is not changed.

kristian-clausal:
Sorry, back from lunch. I see what you mean, I'll try to figure it out.

kristian-clausal:
This is a bigger issue than just "|-", it applies to all combinations of tokens we have... For example

```
{| class="wikitable"
| Testing{|| testing
|}
```
Results in `Testing{` and `testing` cells.

There's a silly little solution (decomposing the token in the parser-function), but it needs more work in cases like this where the token originally parsed was `{|` and then `|` when it was `...{` and `||`, but I think we can use regex lookahead for that.

kristian-clausal:
The linting, co-author stuff, tests, etc. all should be ok, but I'll leave this here overnight in case someone catches something.

xxyzz:
Thanks! The only change I made is remove some `print`.

PR #183, opened 2024-01-12
Fix template expansion recursion too deep error

xxyzz:
[Module:debug/track](https://en.wiktionary.org/wiki/Module:debug/track) expands tracking templates through the Lua frame "expandTemplate" method. The previous code doesn't remove the template name from `Wtp.expand_stack` when template doesn't exist and causes the stack grow too much longer.

The previous code shouldn't skip this line: https://github.com/tatuylonen/wikitextprocessor/blob/8dc20c04ab68fc9c28bcaf751490a14b0fd0fbc8/src/wikitextprocessor/core.py#L1563

Error code: https://github.com/tatuylonen/wikitextprocessor/blob/8dc20c04ab68fc9c28bcaf751490a14b0fd0fbc8/src/wikitextprocessor/core.py#L1365-L1369 

Errors on kaikki.org: https://kaikki.org/dictionary/errors/details-too-deep-recursion-during-template-expansion.html

### DIFF
--- a/src/wikitextprocessor/core.py
+++ b/src/wikitextprocessor/core.py
@@ -1502,7 +1502,8 @@ class Wtp:
                         # print("TEMPLATE_FN {}: {} {} -> {}"
                         #      .format(template_fn, name, ht, repr(t)))
                     if t is None:
-                        ns_id = self.NAMESPACE_DATA["Template"]["id"]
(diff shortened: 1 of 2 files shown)


kristian-clausal:
Good catch! What an annoyingly subtle bug.

PR #185, opened 2024-01-17
Only parse link as plain text if nowiki tag is direct child

### LINKED ISSUE #180: `<nowiki />` tag breaks parsing nodes
Find this error in https://github.com/tatuylonen/wiktextract/pull/453

Example: IPA link in page https://fr.wiktionary.org/wiki/Conjugaison:français/abattre

```
[[Annexe:Prononciation/français|<span>\\kə <nowiki />nu.z‿ɛ.jɔ̃.z‿a.ba.ty\\</span>]]
```

is parsed as plain text because of this code: https://github.com/tatuylonen/wikitextprocessor/blob/0cb88feb77be2e03ba0eac4aca1ea1cb4cba20f8/src/wikitextprocessor/core.py#L598-L600 

can't tell the nowiki tag is inside HTML span tag.

nowiki should only changes the link node to plain text if it's the direct child of the link node, not sure how to fix this bug.

xxyzz:
resolve #180

### DIFF
--- a/src/wikitextprocessor/core.py
+++ b/src/wikitextprocessor/core.py
@@ -599,6 +599,14 @@ class Wtp:
             """Replacement function for links [[...]]."""
             nowiki = MAGIC_NOWIKI_CHAR in m.group(0)
             orig = m.group(1)
+            if MAGIC_NOWIKI_CHAR in orig:
+                # check if nowiki tag is direct child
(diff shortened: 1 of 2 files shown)


kristian-clausal:
I somehow missed this issue, but this seems like an ok fix to a thorny problem.

PR #186, opened 2024-01-17
Fix expand `{{t:=}}` wikitext bug from Russian Wiktionary

xxyzz:
Error page: https://ru.wiktionary.org/wiki/русский


kristian-clausal:
This looks good, thank you for catching this. This would have flown straight over my head!

PR #187, opened 2024-01-22
Remove debug code in luaexec.py

xxyzz:
These debug log code are not needed since the extra arguments bug is fixed by the clone Lua environment feature.

Also ignore extra arguments in Lua API's Python implementation functions, because Lua ignores extra arguments.


xxyzz:
Python functions only need the `*ignore` argument when directly called from Wiktionary Lua code, the newly pushed change removes this argument from my first commit that adds this argument to some indirectly called Python functions in `call_set_functions()`.

PR #190, opened 2024-01-23
Return strip marker from Lua function `preprocess` and `extensionTag`

xxyzz:
Not full implementation of the MediaWiki API, just minimal changes to fix Lua errors in en edition Module:utilities.

GitHub issue tatuylonen/wiktextract#238


kristian-clausal:
This seems like such a thorny issue. Thank you for taking a look :+1:

xxyzz:
I only check the first error page https://en.wiktionary.org/wiki/Reconstruction:Proto-Slavic/divъ

and there are some `rejected heuristic header` debug logs, it seems the code can't parse the declension table properly.

PR #207, opened 2024-02-19
Allow `{` before another curly bracket inside template

xxyzz:
Fixes issue tatuylonen/wiktextract#499

Previous regex matches `{{m|mul|{ }}` but not `{{m|mul{}}`.


kristian-clausal:
This seems correct, I was worrying about matching `{{` inside the template, but that is handled. I'm adding a specific test for that.

PR #212, opened 2024-02-19
Strip newlines from end template arguments when calling

### LINKED ISSUE #208: Strip newline character at the end of unnamed template arguments
Error page: https://ru.wiktionary.org/wiki/adygejski
Error message: https://kaikki.org/ruwiktionary/errors/details-No-declension-pattern-matches--adygejs-jTm2fWw4.html

Wikitext:

```
{{прил pl
|слоги={{по-слогам|a|dy|gej|ski}}|adygejski
}}
```

Lua error:

```
[adygejski (Польский морфологические и синтаксические свойства)](https://kaikki.org/ruwiktionary/All%20languages%20combined/meaning/a/ad/adygejski.html) LUA error in #invoke('inflection adj pl', 'template_decl_auto', 'adygejski\n') parent ('Шаблон:прил pl', {'слоги': 'a<span class="hyph" style="color:lightgreen;">-</span>dy<span class="hyph" style="color:lightgreen;">-</span>gej<span class="hyph" style="color:lightgreen;">-</span>ski', 1: 'adygejski\n'})

[string "inflection adj pl"]:310: No declension pattern matches 'adygejski
'
```

White space characters around unnamed template argument are not removed except the newline character at the end. In this case, "adygejski" should be passed to template ["прил pl"](https://ru.wiktionary.org/wiki/%D0%A8%D0%B0%D0%B1%D0%BB%D0%BE%D0%BD:%D0%BF%D1%80%D0%B8%D0%BB_pl) and "inflection_adj_pl"  lua module not "adygejski\n".

kristian-clausal:
Should fix #208

This stripping is done at the point of calling because removing the newline when in _encode, for example, removes some data we might want to keep for when we use node_to_wikitext and revert the parse tree back to text. Then the newlines would be gone.

Co-authored by: xxyzz

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@
 # Run "make clean" to remove automatically generated files
 ruff:
 	python -m ruff --diff src/wikitextprocessor
+	python -m ruff --diff tests
 test:
 	python -m unittest discover -b -s tests
 test_coverage:
--- a/tests/test_wikiprocess.py
+++ b/tests/test_wikiprocess.py
@@ -4199,7 +4199,17 @@ return export
         )
         self.assertEqual(self.ctx.get_page("mod:title", 828), module_page)
 
-
+    def test_unnamed_template_arg_end_in_newline(self):
+        # https://ru.wiktionary.org/wiki/adygejski
+        # newline at the end of unnamed template argument should be removed
+        self.ctx.add_page("Template:test", 10, "{{{1}}}")
+        self.ctx.start_page("")
+        self.assertEqual(
+            self.ctx.expand("{{test| \n unnamed1 \n}}"), " \n unnamed1 "
+        )
+        self.assertEqual(
+            self.ctx.expand("{{test| \n {{test|unnamed2}} \n}}"), " \n unnamed2 "
+        )
 # XXX Test template_fn
 
 # XXX test post_template_fn
--- a/tests/test_wikiprocess.py
+++ b/tests/test_wikiprocess.py
@@ -4203,12 +4203,29 @@ return export
         # https://ru.wiktionary.org/wiki/adygejski
         # newline at the end of unnamed template argument should be removed
         self.ctx.add_page("Template:test", 10, "{{{1}}}")
+        self.ctx.add_page("Template:testlua", 10, "{{#invoke:test|test|{{{1}}}}}")
+        self.ctx.add_page(
+            "Module:test",
+            828,
+            """
+            local export = {}
+
+            function export.test(frame)
+              return frame:getParent().args[1] .. "-" .. frame.args[1]
+            end
+
+            return export
+            """,)
         self.ctx.start_page("")
         self.assertEqual(
             self.ctx.expand("{{test| \n unnamed1 \n}}"), " \n unnamed1 "
         )
         self.assertEqual(
-            self.ctx.expand("{{test| \n {{test|unnamed2}} \n}}"), " \n unnamed2 "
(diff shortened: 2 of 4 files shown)


kristian-clausal:
This seems to break wiktextract tests in a spectacular way... The tests don't finish!

kristian-clausal:
Actually, I can't get wiktextract tests to finish on other more recent commits. Something's broken on my side...

tests/test_wikiprocess.py
```diff
@@ -4199,7 +4199,17 @@ def test_get_page_with_namespace_prefixes(self):
...
+        # https://ru.wiktionary.org/wiki/adygejski
+        # newline at the end of unnamed template argument should be removed
+        self.ctx.add_page("Template:test", 10, "{{{1}}}")
+        self.ctx.start_page("")
+        self.assertEqual(
+            self.ctx.expand("{{test| \n unnamed1 \n}}"), " \n unnamed1 "
+        )
+        self.assertEqual(
+            self.ctx.expand("{{test| \n {{test|unnamed2}} \n}}"), " \n unnamed2 "
+        )
```
xxyzz, line 4212: This commit changes the template args but not the args used in Lua:

```suggestion
        self.ctx.add_page("Template:test", 10, "{{{1}}}")
        self.ctx.add_page("Template:testlua", 10, "{{#invoke:test|test|{{{1}}}}}")
        self.ctx.add_page(
            "Module:test",
            828,
            """
            local export = {}

            function export.test(frame)
              return frame:getParent().args[1] .. "-" .. frame.args[1]
            end

            return export
            """,)
        self.ctx.start_page("")
        self.assertEqual(
            self.ctx.expand("{{test| \n unnamed1 \n}}"), " \n unnamed1 "
        )
        self.assertEqual(
            self.ctx.expand("{{test| \n {{test|unnamed2}} \n}}"),
            " \n unnamed2 ",
        )
        self.assertEqual(
            self.ctx.expand("{{testlua| \n unnamed3 \n}}"), " \n unnamed3 - \n unnamed3"
        )
```

If the last assert passed then the Lua error in ru edition page should be fixed.

kristian-clausal:
Oh, that's a conundrum... Merged the tests into the pull request (I think? surely this doesn't merge into main).

xxyzz:
My first comment was missing a white space at the end, you were faster than me...

But that's OK, the strip newline code seems still not at the right place.

xxyzz:
How about add another bool argument(default to `False`) to `vbar_split` and pass `True` from `repl_templ` then only remove the last `\n` if the argument is `True`? Only two tests failed with this change.

kristian-clausal:
I'm trying to see if I can figure out a perfect place to remove the newline. We really wish to keep the parse tree reversion as is.

Makefile
```diff
@@ -2,6 +2,7 @@
 # Run "make clean" to remove automatically generated files
 ruff:
 	python -m ruff --diff src/wikitextprocessor
+	python -m ruff --diff tests
```
xxyzz, line 5: I'd recommend `ruff format .`, this command will format the files for you.

xxyzz:
Thanks you! The Lua error is fixed.

Makefile
kristian-clausal: I'm formatting by hand inside the editor (Helix, `:format` command), because I had trouble with venv breaking ruff installs when I was in the wrong environment and basically destroying files when using auto-format on save. Just being careful. The command should probably be just named "make ruffcheck" or something...

PR #214, opened 2024-02-21
Fix bug in _sandbox_phase2/make_frame/extensionTag

kristian-clausal:
`args[0]["args"]` is a full Lua table, which is not necessarily all string values. This caused at least some Lua errors when html.escape assumed `v` was a string; `v.replace()` would panic.

### DIFF
--- a/src/wikitextprocessor/luaexec.py
+++ b/src/wikitextprocessor/luaexec.py
@@ -481,7 +481,7 @@ def call_lua_sandbox(
                 attrs2 = list(
                     v
                     if isinstance(k, int)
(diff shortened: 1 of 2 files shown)


xxyzz:
Could you add a test?

This error would be much easier to find if lua's `pcall` doesn't remove the error stack log.

kristian-clausal:
Tried to make sure "HTML" tag attributes can't be anything else but strings (I am 99% sure they're always strings). `<maplink>` attribute data types are definitely strings, so they must be converted in Wikimedia, which should be the correct way to do this on our side, too.

PR #215, opened 2024-02-21
Chaining methods in mw.html should probably return self

kristian-clausal:
Fix this bug:

```lua
		local htmlTitle = mw.html.create('span')
			:attr{ id = 'coordinates' }
			:addClass( displayinline and 'noprint' or nil )
			:wikitext( "" )
```

Our implementation would break after :addClass because it would return nil, when it should have returned `self`, so that :wikitext() would not try to 'index' a nil value. Even if the mw.html method doesn't succeed, it should return `self` for this chaining; I checked the scribunto source, and none of the methods returned without returning something or `self`.

### DIFF
--- a/src/wikitextprocessor/core.py
+++ b/src/wikitextprocessor/core.py
@@ -205,6 +205,7 @@ class Wtp:
         "lua_frame_stack",
         "project",  # "wiktionary" or "wikipedia"
         "strip_marker_cache",
+        "ALLOWED_EXTENSION_TAGS",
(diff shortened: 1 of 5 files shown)


kristian-clausal:
I assume it's for security reasons. Because Tatu's done it this way, I assume it has some relevance.

xxyzz:
That's unlikely, the `tag` variable is not used to build executable string like a SQL query.

PR #217, opened 2024-02-22
Move HTML tag data to Wtp context instead of global

kristian-clausal:
Because we want to allow parsing HTML-like extension tags, like `<maplink>` data, we need to add them to the HTML tag data we have, but changing ALLOWED_HTML_TAGS and messing around with capitalised globals doesn't seem the way to go.

ALLOWED_HTML_TAGS will stay as the base on which everything is built, but what should be generally used in the future instead is Wtp.allowed_html_tags, which should also include HTMLTagData entries for extra tags.

Because a bunch of other globals rely on ALLOWED_HTML_TAGS, these also need to be brough back inside the context object, generated in Wtp.__init__().

There was one place with ALLOWED_HTML_TAGS left outside of the context, in `to_wikitext` where a parse tree is turned back into text; here, it was used only to check if a lone tag (`<tag>` without an `</tag>` and without children inside the element) should be outputted with a self-closing slash (`<tag />`) or not; in practice, this should not matter.

The code in Wikitextprocessor doesn't *handle* the HTML element, so currently adding capability to parse a `<maplink>` element doesn't function correctly because it outputs the contents when it apparently shouldn't.

The handling of the extension tags should probably be at a higher level than the parser.

### DIFF
--- a/src/wikitextprocessor/core.py
+++ b/src/wikitextprocessor/core.py
@@ -49,18 +49,15 @@ from .parser import (
     GeneralNode,
     WikiNode,
     parse_encoded,
+    set_html_tag_data,
+    set_inside_html_tags_re,
 )
 from .parserfns import PARSER_FUNCTIONS, call_parser_function
-from .wikihtml import ALLOWED_HTML_TAGS
+from .wikihtml import ALLOWED_HTML_TAGS, HTMLTagData
 
 if TYPE_CHECKING:
     from lupa.lua51 import LuaNumber, LuaRuntime, _LuaTable
 
-
-# Set of HTML tags that need an explicit end tag.
-PAIRED_HTML_TAGS: set[str] = set(
-    k for k, v in ALLOWED_HTML_TAGS.items() if not v.get("no-end-tag")
-)
 
 NamespaceDataEntry = TypedDict(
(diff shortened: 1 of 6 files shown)


xxyzz:
I still think we shouldn't make rules don't exist on MediaWiki, who wants to check which extension creates what tags and pass this data to `Wtp.__init__()`?

kristian-clausal:
> I still think we shouldn't make rules don't exist on MediaWiki, who wants to check which extension creates what tags and pass this data to `Wtp.__init__()`?
> 
> And because the `content` is not checked, any tags not allowed can be put in the `content` argument:
> 
> ```lua
> frame:extensionTag("a", "<bbbb>")
> ```

I don't understand what you're saying.

xxyzz:
I mean the tag check code is not effectively, any tag can be created by putting them in the `content` or `args` argument of the `#tag` parser function or `frame:extensionTag()`. And most importantly, MediaWiki allows any tag to be created.

kristian-clausal:
ALLOWED_HTML_TAGS is the way we check if a tag is allowed. Our code (currently) does not allow extension tags that aren't there, *and* the commit in the other PR I made wasn't enough to make them work. This all is necessary to get extension tags to work.

PR #222, opened 2024-02-23
Implement partial #property parserfn

kristian-clausal:
Currently, #statements return some raw data without processing happening, which is what we would want
.#property to be, although #property also changes the data.

For example, coordinates are formatted, and should not be in the form 'Point(2.294479 48.858296)'.

We might need to implement all of these if we want to be 'correct'...

This commit just takes the code from #statements, puts it in .#property where it fits 'better' and then has #statements called property_fn; same output of 'raw' data, which is hopefully good enough for most purposes.

### DIFF
--- a/src/wikitextprocessor/parserfns.py
+++ b/src/wikitextprocessor/parserfns.py
@@ -1359,6 +1359,17 @@ def statements_fn(
     wtp: "Wtp", fn_name: str, args: list[str], expander: Callable[[str], str]
 ) -> str:
     # https://www.wikidata.org/wiki/Wikidata:How_to_use_data_on_Wikimedia_projects
+    # XXX? This implementation doesn't implement the fancy things #statements
+    # generates, like links or images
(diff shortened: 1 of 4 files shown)


xxyzz:
The interactive map(the `maplink` tag) is for javascript in browsers, we could create the same format as the wikidata extension but we don't really need it because we'll never process it.

xxyzz:
The link format is a bit different. For the "External IDs" IMDb example in the [wikidata document](https://www.wikidata.org/wiki/Wikidata:How_to_use_data_on_Wikimedia_projects),  I think wikidata uses the the "[formatter URL](https://www.wikidata.org/wiki/Property:P1630)" property to format the IMDB id. Unless we have to use the link, we don't need to format the ids because the these links still display as their id text.

I should add the wikidata data type, my previous code only format one "Point in time" type property.

I start to think we could just call mediawiki api to get these parser function's output, there is not much difference(we already call wikidata's api to get data) and easier to implement.

PR #227, opened 2024-02-27
Allow slash in HTML attribute value

xxyzz:
Fix parse exception page de edition page "axitiosus", and use `raise Exception` so the `token` variable value could be included in the error message.

Error page: https://kaikki.org/dewiktionary/errors/details-----EXCEPTION-while-parsing-page--axit-Ob46aw13.html

### DIFF
--- a/src/wikitextprocessor/parser.py
+++ b/src/wikitextprocessor/parser.py
@@ -1812,7 +1812,7 @@ def tag_fn(ctx: "Wtp", token: str) -> None:
     # Try to parse it as a start tag
     m = re.match(
         r"""<([-a-zA-Z0-9]+)\s*((\b[-a-zA-Z0-9:]+(\s*=\s*("[^"]*"|"""
-        r"""'[^']*'|[^ \t\n"'`=<>]*))?\s*)*)(/?)\s*>""",
+        r"""'[^']*'|[^ \t\n"'`=<>]*))?\s*)*)(/?)>""",
         token,
     )
     if m:
--- a/src/wikitextprocessor/parser.py
+++ b/src/wikitextprocessor/parser.py
@@ -1812,7 +1812,7 @@ def tag_fn(ctx: "Wtp", token: str) -> None:
     # Try to parse it as a start tag
     m = re.match(
         r"""<([-a-zA-Z0-9]+)\s*((\b[-a-zA-Z0-9:]+(\s*=\s*("[^"]*"|"""
-        r"""'[^']*'|[^ \t\n"'`=<>/]*))?\s*)*)(/?)\s*>""",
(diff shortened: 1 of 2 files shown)


src/wikitextprocessor/parser.py
```diff
@@ -1812,7 +1812,7 @@ def tag_fn(ctx: "Wtp", token: str) -> None:
     # Try to parse it as a start tag
     m = re.match(
         r"""<([-a-zA-Z0-9]+)\s*((\b[-a-zA-Z0-9:]+(\s*=\s*("[^"]*"|"""
-        r"""'[^']*'|[^ \t\n"'`=<>/]*))?\s*)*)(/?)\s*>""",
+        r"""'[^']*'|[^ \t\n"'`=<>]*))?\s*)*)(/?)\s*>""",
```
kristian-clausal, line 1815: Yeah, this is correct (we might be able to remove other symbols too from the prohibited list, but one step at a time). Tatu is a bit overzealous with being careful, and probably didn't look at the HTML syntax specs. Then again, what is and isn't allowed in HTML has changed over the years so much and been so lax that I don't blame him.

Because "/>" is a token that can't have whitespace inside of it, we can change `(/?)\s*` to `(/?)`. I've been trying to spot these when they come along, and I've removed some of them, but I bet there's still more. Basically, it seems that the HTML tag syntax is that each of `<tagname`, `</endtagname` and `/>` is one token without whitespace inside of it, so `<tagname >`, `</endtagname   >`, `<tagname   />` is allowed, but not `<   tagname>`, `</  endtagname>` or `<tagname /   >`.

src/wikitextprocessor/parser.py
```diff
@@ -1919,8 +1919,7 @@ def tag_fn(ctx: "Wtp", token: str) -> None:
     else:
         m = re.match(r"</([-a-zA-Z0-9]+)\s*>", token)
         if m is None:
-            print("Could not match end tag token: {!r}".format(token))
-            assert False
+            raise Exception("Could not match end tag token: {!r}".format(token))
```
kristian-clausal, line 1922: Oh, I haven't even realized this; yeah, of course raising an Exception is better than assert False. I'll have to keep this in mind. Thanks for pointing this out.

tests/test_parser.py
```diff
@@ -2823,9 +2823,7 @@ def test_extension_tags(self):
         extension_tags = {
             "foo": {"parents": ["phrasing"], "content": ["phrasing"]},
         }
-        self.ctx = Wtp(
-            extension_tags=extension_tags,
-        )
+        self.ctx.allowed_html_tags.update(extension_tags)
```
kristian-clausal, line 2826: Thanks for the correction, I didn't even think about this.

kristian-clausal (approved):

Issue #225, opened 2024-02-23
Template {{Voir homonymes|....}} is misinterpreted


LeMoussel:
On Wikipedia [Aisne (département)](https://fr.wikipedia.org/wiki/Aisne_(d%C3%A9partement)) page, the wikitext of _Pour les articles homonymes, voir [Aisne](https://fr.wikipedia.org/wiki/Aisne)_ is misinterpreted.

The Wikitext template is: `{{Voir homonymes|Aisne}}`

Output Text:
`class=noviewer}
Erreur : Paramètre |texte= manquant
| texte = Pour les articles homonymes, voir Aisne.
}}`

Test code:
```python
from unittest import TestCase

from wikitextprocessor import Wtp
from wiktextract.page import clean_node
from wiktextract.wxr_context import WiktextractContext
from wiktextract.config import WiktionaryConfig


class TestHomonymes(TestCase):
    def setUp(self):
        self.wxr = WiktextractContext(
            wtp = Wtp(
                db_path="fr-wiki-latest.db",
                lang_code="fr",
                project="wikipedia",
            ),
            config=WiktionaryConfig()
        )

    def tearDown(self):
        self.wxr.wtp.close_db_conn()

    def test_homonymes(self):
        self.wxr.wtp.start_page("Test-Homonymes")
        tree = self.wxr.wtp.parse(text="{{Voir homonymes|Aisne}}", expand_all=True)
        text = clean_node(
            wxr=self.wxr,
            sense_data={},
            wikinode=tree,
        )
        self.assertGreater(len(text), 0)
```

kristian-clausal:
I am pretty sure there's a syntax error in 

https://fr.wikipedia.org/w/index.php?title=Mod%C3%A8le:M%C3%A9ta_bandeau_de_note&action=edit

```
<includeonly>{{Méta bandeau
| forme = note
| niveau = aucun
| class = hatnote {{{class|}}}
| id = {{{id|}}}
| style = 
| icône = {{#switch: {{{icône|}}}
| confusion   = [[Fichier:Confusion colour.svg|20px|link={{{link|Aide:Homonymie}}}|alt={{{alt|Page d’aide sur l’homonymie}}}|class=noviewer]]
| homonymie   = [[Fichier:Logo disambig.svg|20px|link={{{link|Aide:Homonymie}}}|alt={{{alt|Page d’aide sur l’homonymie}}}|class=noviewer]]
[...]

PR #228, opened 2024-02-28
Changes to parser regex

kristian-clausal:
Partial fix to #225

We can't get 100% correspondence with how Wikimedia parses things because our parser works on different principles, but here's a fix for an edge case. I found a few other edge-cases (added commented-out tests for them) that I don't think we can fix.

The regexes for the parsing-loop in core.py where we turn links, external links, template arguments and templates into magical characters to be stored out of the way while doing inside-out parsing are now outside the function def itself in global space.

The link regex has been changed to accept (in some cases) stray square brackets inside the link itself (but not double brackets or paired brackets).

The link regex also can now contain curly braces, etc. because we handle checking whether the link contains templates etc. in the regex handling function passed into re.sub:

The link substitution function (used to handle the regex matches we come across that match this) has been modified to look for a combination of all the possible regexes that were moved out into global namespace. So if the link regex lets something through (like a template), it will spot it at that point and then returns the original string (which means prev == current, and the loops terminates to handle other stuff instead and returns later to handle the link, if applicable).

### DIFF
--- a/src/wikitextprocessor/core.py
+++ b/src/wikitextprocessor/core.py
@@ -157,7 +157,7 @@ class BegLineDisableManager:
             self.ctx.begline_enabled = True
 
 
-links = (
+LINKS = (
     # `[[something<abcd>]]`
     # XXX this regex seems to be too complex,
     # could you replace it with just [^][{}]*?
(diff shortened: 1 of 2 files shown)


xxyzz:
This is so scary, my only suggestion is use upper case name for global variables...

kristian-clausal:
I know! Regexes.

PR #235, opened 2024-03-05
Getbadge

kristian-clausal:
Incomplete getBadge() implementation; return an empty table. We don't *really* need to know what badges a wikipedia page has for parsing, and this will fix issues.

getBadge was introduces in 2022~2023, so it is not strange it was unimplemented.

Afaict, the code doesn't call anything like Wikidata for the badge stuff, it's internal stuff (but with globalsiteid??), and I'm not even sure the dump data has badge data at all.

### DIFF
--- a/src/wikitextprocessor/lua/mw_wikibase.lua
+++ b/src/wikitextprocessor/lua/mw_wikibase.lua
@@ -23,7 +23,7 @@ function mw_wikibase.getEntityUrl(id)
    return nil
 end
 
-function mw_wikibase.getBadge(id, globalSiteId)
+function mw_wikibase.getBadges(id, globalSiteId)
    return {}
 end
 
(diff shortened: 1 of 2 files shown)


src/wikitextprocessor/lua/mw_wikibase.lua
```diff
@@ -23,6 +23,10 @@ function mw_wikibase.getEntityUrl(id)
    return nil
 end
 
+function mw_wikibase.getBadge(id, globalSiteId)
```
xxyzz, line 26: Should be `getBadges`, other than this typos the commit LGTM.

kristian-clausal: Oh man, yeah, the plural -s strikes again. Also the same typo in the 'test'.

PR #238, opened 2024-03-07
Some regex pattern object changes

xxyzz:
Turns out https://github.com/tatuylonen/wikitextprocessor/commit/c25029df514c7e34820782476e6d297150081eaf makes the code slow(the `Wtp.preprocess_text()` function)... but `Wtp.process_text` is also called before `Wtp._encode()` in `Wtp.expand()`. This pr remedies a bit, after apply this pr the code is still twice slow as the code before c25029df514c7e34820782476e6d297150081eaf but faster than master.

### DIFF
--- a/src/wikitextprocessor/__init__.py
+++ b/src/wikitextprocessor/__init__.py
@@ -1,4 +1,5 @@
-from .core import MAGIC_FIRST, MAGIC_LAST, Page, Wtp
+from .common import MAGIC_FIRST, MAGIC_LAST
+from .core import Page, Wtp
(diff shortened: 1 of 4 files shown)


kristian-clausal (approved):
I didn't realize the cache-checking in `re` was this impactful, even with pre-compiled patterns! Also, I didn't realize you could just call a method on a pattern to replicate these `re` functions, I will have to keep in mind. All of these changes seem 100% straightforward and correct to me!

kristian-clausal:
Doing these smaller optimizations doesn't go to waste, though, so it was good to get them done!

PR #244, opened 2024-03-08
Fix: mistaken implementation of mw.text.decode

kristian-clausal:
Our implementation of mw.text.decode was broken due to a misreading of
https://www.mediawiki.org/wiki/Extension:Scribunto/Lua_reference_manual#mw.text.decode

Here it is said that

> If boolean decodeNamedEntities is omitted or false,
> the only named entities recognized are `&lt;` (<),
> `&gt;` (>), `&amp;` (&), `&quot;` (") and `&nbsp;`
> (the non-breaking space, U+00A0).
> Otherwise, the list of HTML5 named entities to
> recognize is loaded from PHP's get_html_translation_table function.

Tatu read this as "if decodeNamedEntities is omitted or false, decode only these values", but it turns out that `decodeNamedEntities` is only required to decode *name entities* that aren't in this list; all number values are still decoded. I have no idea why this is like this, and I had the same misunderstanding when first looking at the documentation, but the Scribunto source code was clearer.

This was causing a ton of weird Lua errors in etymology templates, because there are bits of code that rely on the presence of a *asterisk, used in historical linguistics to denote a form of a word that is only reconstructed and not attested.

So when these entities were encoded into HTML entities (for some reason) and then the decode failed because our implementation was wonky, it broke the etymology template and it took me the hole day of hunt-and-peck print debugging to finally find where the fault was.

### DIFF
--- a/src/wikitextprocessor/luaexec.py
+++ b/src/wikitextprocessor/luaexec.py
@@ -84,33 +84,19 @@ def lua_loader(ctx: "Wtp", modname: str) -> Optional[str]:
     return data
 
 
+html_entities_re = re.compile(r"&(lt|gt|amp|quot|nbsp|#[xX]?[0-9A-Fa-f]+);")
+
+
+def replace_specific_entities(m: re.Match) -> str:
+    return html.unescape(m.group(0))
+
+
 def mw_text_decode(text: str, decodeNamedEntities: bool) -> str:
     """Implements the mw.text.decode function for Lua code."""
     if decodeNamedEntities:
         return html.unescape(text)
 
-    # Otherwise decode only selected entities
-    parts: list[str] = []
-    pos = 0
-    for m in re.finditer(r"&(lt|gt|amp|quot|nbsp);", text):
-        if pos < m.start():
-            parts.append(text[pos : m.start()])
-        pos = m.end()
-        tag = m.group(1)
-        if tag == "lt":
-            parts.append("<")
-        elif tag == "gt":
-            parts.append(">")
-        elif tag == "amp":
-            parts.append("&")
-        elif tag == "quot":
-            parts.append('"')
-        elif tag == "nbsp":
-            parts.append("\xa0")
-        else:
-            assert False
-    parts.append(text[pos:])
-    return "".join(parts)
+    return html_entities_re.sub(replace_specific_entities, text)
 
 
 def mw_text_encode(text: str, charset: str) -> str:
--- a/tests/test_wikiprocess.py
+++ b/tests/test_wikiprocess.py
@@ -4099,9 +4099,7 @@ return export
         )
 
     def test_gsub5(self):
-        self.scribunto(
-            "42", """return string.gsub("x2A", "x%x+", "42");"""
-        )
+        self.scribunto("42", """return string.gsub("x2A", "x%x+", "42");""")
 
     def test_title_1_colon_e(self) -> None:
         self.scribunto("1:e", "return mw.title.new('1:e').text")


xxyzz:
Should add the Scribunto code link and some tests, the description of the problem is confusing...

src/wikitextprocessor/luaexec.py
```diff
@@ -84,33 +84,19 @@ def lua_loader(ctx: "Wtp", modname: str) -> Optional[str]:
     return data
 
 
+html_entities_re = re.compile(r"&(lt|gt|amp|quot|nbsp|#[xX]?[0-9A-Fa-f]+);")
```
xxyzz, line 87: `A-Fa-f` must be a typo, I fixed the pattern in #247.

I guess the pr is for the Lua error in page https://en.wiktionary.org/wiki/galar, wikitext: `{{inh|ga|ine-pro|[[*ǵʰelH-]]ro-}}`

lua error: 

```
[galar (Scottish Gaelic top-level)](https://kaikki.org/dictionary/All%20languages%20combined/meaning/g/ga/galar.html) LUA error in #invoke('etymology/templates/inherited', 'inherited') parent ('Template:inherited', {1: 'gd', 2: 'ine-pro', 3: '[[*ǵʰelH-]]ro-'})

[string "Module:links"]:180: The specified language Proto-Indo-European is unattested, while the given word is not marked with '*' to indicate that it is reconstructed.

Path: galar > inh > #invoke > #invoke
```

but I couldn't find the relation between this pr and the lua error.

kristian-clausal:
A-Fa-f is not a typo, it's for hexadecimal numerals!

This is not scribunto HTML entities, these are literal *HTML*-entities, so the 'X' is allowed!

Revert the changes.

kristian-clausal:
The path of the error is that HTML entities for asterisks (`&#xsomething`) were being generated in the module code, but our decoder was not able to decode any HTML-entity numerals because it was written wrong. This lead to parts of the module code to split the string it got on the `#` and only use the `&` instead of `*protoword`, which broke it.

kristian-clausal:
Looking at the mw.text.decode implementation's code for Scribunto, you are correct that it doesn't accept capital 'X'. The module code has some patterns that have `[xX]?` in this location, but that's wrong on their part.

xxyzz:
I only looked the Scribunto code and didn't realize it's for hex string, I guess "a-z" also works so the Scribunto developers keep the wrong pattern.

kristian-clausal:
Our pattern is just more discerning, theirs is a Lua pattern which has to be simpler.

xxyzz:
I'll create a pr to restore your previous pattern. And could you explain how this pr fixes the error in page [galar](https://en.wiktionary.org/wiki/galar)?

kristian-clausal:
"The path of the error is that HTML entities for asterisks (&#xsomething) were being generated in the module code, but our decoder was not able to decode any HTML-entity numerals because it was written wrong. This lead to parts of the module code to split the string it got on the # and only use the & instead of *protoword, which broke it."

The etymology modules turn `*` into HTML entities on purpose, then that HTML entity failed to decode because it was encoded as a numeral, and the string was later split in the '#' leaving a string with only '&' instead of a string like '*protoword'. This is what happened. The decoding failing meant that a string was broken, which caused errors.

xxyzz:
Sorry I missed that comment. IIUC, encode and decode of the `*` character in the "inh" template argument happens somewhere in Wiktionary 's Lua module code but not relevant to our encode and decode arguments Python code? I think I made some changes to the encode and decode template arguments sometime ago.

kristian-clausal:
Our implementation of decode was broken because it would not decode numerals when it should (=always).

The argument given to decode() is about whether you only allow "nbsp" and the other listed named entities to be decoded, or also *other* named entities from a larger list.  For example, if `decodeNamedEntities` is set to true, *then* is when decode() will decode `&plusmn;`; but otherwise, it will not, so when `decodeNamedEntities` is not set or false it wouldn't.

But the original implementation would not decode numeral entities when decodeNamedEntities was not set because of misunderstanding the documentation. Numerals should always be decoded.

xxyzz:
I mean it's not our code that convert the argument like here: https://github.com/tatuylonen/wikitextprocessor/blob/5f6ddbb846f305add1052e855e2b8fd380fde2f7/src/wikitextprocessor/luaexec.py#L478, at first I though it might be code like this causes the error but I guess this is irrelevant.

kristian-clausal:
Yeah, the issue was that Wiktionary-side modules converted stuff (including '*') into HTML-entities; the '*' is weird, but it's done in a function called 'risky' something or other, so maybe there's a reason.

kristian-clausal:
While on my lunch-break, I realized why the pattern in Scribunto, `(&(#?x?)([a-zA-Z0-9]+);)` doesn't break when it has 'X' instead of 'x'... The [a-zA-Z] also includes 'X'. The `x?` doesn't do anything in the pattern, :smile: .

PR #246, opened 2024-03-08
Implement localdow parserfns 

LeMoussel:
Implement
- `{{LOCALDOW}}` “Local” rank of current weekday (0–6)


xxyzz:
Your branch was created from your last pr's branch so it also contains commits in that pr, I rebase to main branch and fix the return value.

PR #247, opened 2024-03-09
Fix HTML regex pattern and add a test for `mw.text.decode()`

xxyzz:
Upper case `X` before HTML entity number is not supported in Scribunto, so we could only use the lower case `x`.


kristian-clausal:
These changes are incorrect! a-fA-F is for hexadecimal numbers! The uppercase 'X' is taken directly from module code!

PR #250, opened 2024-03-11
Fix misuse of `Match.start()` causes template expanded as text bug

xxyzz:
`m.start(1)` returns the location in the text passed to `re.sub()` not the location in the matched group text.

Fix https://github.com/tatuylonen/wiktextract/issues/533

### DIFF
--- a/src/wikitextprocessor/core.py
+++ b/src/wikitextprocessor/core.py
@@ -660,10 +660,9 @@ class Wtp:
             functions."""
             whole_match = m.group(0)
             nowiki = False
-            if (
-                MAGIC_NOWIKI_CHAR
-                in whole_match[: m.start(1)] + whole_match[m.end(1) :]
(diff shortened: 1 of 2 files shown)


xxyzz:
If I use `args[0].strip() == ""` at https://github.com/tatuylonen/wikitextprocessor/blob/f2fbe0495f975c5b28b16aa1a8ba8640217efaa9/src/wikitextprocessor/core.py#L669, then test [test_left_curly_bracket_in_template2](https://github.com/tatuylonen/wikitextprocessor/blob/f2fbe0495f975c5b28b16aa1a8ba8640217efaa9/tests/test_parser.py#L2802) will fail. I'll ignore this for now...

kristian-clausal:
Hopefully 'llave' won't break again. Thanks for looking at this.

PR #252, opened 2024-03-12
Determine a HTML tag is self-closing if it ends with "/>"

xxyzz:
Fixes tatuylonen/wiktextract#535

It works for now, but we'll have more errors with this regex in the future, only using a real HTML parser could fix them.


kristian-clausal:
This seems like a slam dunk.

kristian-clausal (commented):
I do wonder how much more efficient this is (how much resources do regex groups 'use'?), but this is definitely clearer.

xxyzz:
Probably negligible in most cases. Speaking of performance, extract fr edition time drops back to 40 minutes, maybe some commits after #238 improved the speed.

PR #253, opened 2024-03-12
Add `file_aliases` parameter to Wtp

kristian-clausal:
Some wikiprojects have translated syntax for
`[[File` / `[[Image` image/file display links, like French wikipedia `Fichier`.

In Wiktextract clean_value, links with `File` and
`Image` are basically returned as empty strings, so we'll continue doing that for now; might change things for `class=noviewer` which inlines an image, and which might require handling of alt texts.


xxyzz:
Namespace alias data contain this "Fichier" value, I should update the "get_namespaces.py" file and namespace json files.

Issue #226, opened 2024-02-26
Checklist-1 for existing errors.


LeMoussel:
Attached CSV file: [wiki_errors.csv](https://github.com/tatuylonen/wikitextprocessor/files/14405399/wiki_errors.csv), listing, by Wikipedia article title, other errors than those indicated in the issues #225, #224, #223, #220 & #216:

In summary, there are the following errors:

- [x]  LUA error in #invoke('Cat\u00e9gorisation badges', 'badgesCategories') - Corrected
- [x] unimplemented parserfn #property  - Corrected
- [x] LUA error in #invoke('Composition Division de France', 'mod\u00e8le')  - Corrected
- [x] LUA error in #invoke('Jumelages', 'tableauDesJumelages')  - Corrected
- [x] unimplemented parserfn PROTECTIONLEVEL  - Corrected

LeMoussel:
For `unimplemented parserfn PROTECTIONLEVEL`,  I suggest this correction in [parsefn.py](https://github.com/tatuylonen/wikitextprocessor/blob/main/src/wikitextprocessor/parserfns.py#L1499)
```python
....
    "PROTECTIONLEVEL": protectionlevel_fn, #unimplemented_fn,
....

def protectionlevel_fn(
    ctx: "Wtp", fn_name: str, args: list[str], expander: Callable[[str], str]
) -> str:
    """Implements the PROTECTIONLEVEL magic word."""
    # Returns an empty string to indicate that the page is not protected."""
    return ""
```

kristian-clausal:
If you don't want to make a pull request, I'll implement that. It seems the most sensible approach here, yeah.

LeMoussel:
I don't know GitHub well to make a pull request. I'll let you do it. Thanks.

kristian-clausal:
Pushed a commit to the #filepath PR (might as well lump them together).

kristian-clausal:
@LeMoussel I've not gone through and committed fixes to most issues. The CSV had two new issues, `mw.ext.data.get` not being implemented (it's an extension that isn't used in Wiktionary) and `getBadges` not being implemented (it's a new function introduced in 2022~2023).

`PROTECTIONLEVEL`, `#property` should also be handled. If you could check out all of these issues and see if they work on your end now, that would be grand.

LeMoussel:
First of all, let me congratulate you both. You do a hell of a job! THANKS.

[...]

PR #256, opened 2024-03-14
Implement mw.language:formatDate()

kristian-clausal:
Issue #226 case 'Durée'

mw.language:formatDate() was only partially implemented.

However, the same kind of code needed there has already been (mostly?) implemented in the `{{#time:...}}` parser function in parserfns, so we can reuse code from there by refactoring the meaningful bits out and making a new mw_language_format_date_python function to be called from Lua code that will do most of the things needed here.

I spent too much time trying to roll my own implementation... before realizing there already was one in parserfns!!


LeMoussel:
Resolved

PR #260, opened 2024-03-25
Implement `mw.wikibase.getEntity` Lua API

xxyzz:
Cache entity data JSON string to SQLite file.

Hard code the language code and site id as "en" and "enwiki", should use values in the `Wtp` object, not sure how to setup the object like the MediaWiki code(`Entity.setupInterface`).

`mw.wikibase.entity:formatPropertyValues` and
`mw.wikibase.entity:formatStatements` are not implemented, they use php code to format values.

Some methods of the entity object call not implemented functions of the `mw.wikibase` table, errors are expected.


kristian-clausal:
Thank you very much for taking this on!

PR #267, opened 2024-04-10
Change to link detection regex

### LINKED ISSUE #266: Can't parse link nodes contain newline character
Page: https://en.wiktionary.org/wiki/forswat
Simplified Wikitext: `[[a|b\nc]]`
Error message: https://kaikki.org/dictionary/All%20languages%20combined/errors/details--2--is-an-alias-of--year---cannot-spec-Q6~yILHj.html

The links regex at here https://github.com/tatuylonen/wikitextprocessor/blob/cdd76b208685d2e040e03a95a9ecde8e89390c68/src/wikitextprocessor/core.py#L160

can't match the `[[a|b\nc]]` link, @kristian-clausal could you please take a look of the regex? I'm not dare to change this pattern...

kristian-clausal:
Should fix #266

The previous version did not allow for links with newlines in the text portion of the link:

```
[[this link|

should be

accepted]]
```

Newlines are not allowed in the link data portion ('this link').

Because of the negative lookahead stuff to detect `[[`, `]]` and `[`, `]` pairs inside the links, it's a real monster of a regex that's hard to read.

I also found a minor bug that meant a part of the regex was basically disabled, so I just removed it (no ^ to negate \n in [\n]).

Also added a negative lookbehind: a link starting with `[[[` is parsed as text.

### DIFF
--- a/src/wikitextprocessor/core.py
+++ b/src/wikitextprocessor/core.py
@@ -161,16 +161,34 @@ LINKS = (
     # `[[something<abcd>]]`
     # XXX this regex seems to be too complex,
     # could you replace it with just [^][{}]*?
-    r"\["
+    r"(?<!\[)"
+    + r"\["
     + MAGIC_NOWIKI_CHAR
     + r"?\[("
-    # I spent a lot of willpower on this monstrosity.
-    + r"((?!\]\])[^[\n])*(?!\[[\n]+\])((?!\[\[)[^]\n])+"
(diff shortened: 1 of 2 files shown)


xxyzz:
This feels like a parser would do, look around neighbor tokens to decide how to parse tokens. I wonder which is more painful, use a better parser(maybe [mwparserfromhell](https://github.com/earwig/mwparserfromhell)?) or add more regex...

PR #270, opened 2024-04-17
Remove unnecessary warning and logging setting code

xxyzz:
- DEBUG logging level is only set to our own `Logger` obejct in wiktextract code and won't affect other packages: tatuylonen/wiktextract#591
- I guess the warning filter was added for the "dateparser" package, this was fixed two years ago(dateparser GH issue 1013)

### DIFF
--- a/src/wikitextprocessor/parserfns.py
+++ b/src/wikitextprocessor/parserfns.py
@@ -3,7 +3,6 @@
 # Copyright (c) 2020-2022 Tatu Ylonen.  See file LICENSE and https://ylonen.org
 
 import html
-import logging
 import math
 import re
 import urllib.parse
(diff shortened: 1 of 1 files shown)


xxyzz:
Perhaps we should consider use Python's `logging` library in wikitextprocessor? It could print error messages to error output and save logs to a file so we don't need to worry logs take too much memory. And we also might use other features from `logging`.

kristian-clausal:
Let's introduce new imports only when necessary. Wikitextprocessor has worked well without logging up till now, and we have our own error logging system. If there is a pressing need to print logging stuff and no way to do it, then it should be done, but not prematurely.

xxyzz:
The main problem with our current `Wtp.error()` methods is they save data in memory not how they are printed, and the recent check JSON code adds more JSON data to the log messages which exacerbates the problem. The server currently has enough RAM but IMO this could be improved.

Issue #258, opened 2024-03-19
Undeclared variable assignments from modules without require ('strict');


olsaarik:
I'm processing a recent English Wikipedia dump and getting `assign to undeclared variable` errors from modules that don't have a `require ('strict');` in them. Here's stripped down code to replicate this:

```
from wikitextprocessor import Wtp
wtp = Wtp(db_path="enwiki-20240201-wtp.db", lang_code="en", project="wikipedia")
process_dump(wtp, "enwiki-20240201-pages-articles-multistream.xml.bz2", {0, 10, 828}) # main, Template and Module namespaces
for page in wtp.get_all_pages():
    if page.redirect_to != None:
        continue
    wtp.start_page(page.title)
    wtp.expand(page.body)
    break
```

And here's the errors I'm seeing:

```
Anarchism: ERROR: LUA error in #invoke('Protection banner', 'main') parent ('Template:Pp-semi-indef', {}) at ['Anarchism', 'pp-semi-indef', '#invoke', '#invoke']
[string "Module:Effective protection level"]:63: attempt to index field 'TitleBlacklist' (a nil value)
Anarchism: WARNING: invalid attribute format '20x20px\n ' missing name at ['Anarchism', 'good article', 'Main other', 'ARGVAL-1', 'Top icon', '#tag', '#tag']
Anarchism: WARNING: invalid attribute format '\n ' missing name at ['Anarchism', 'good article', 'Main other', 'ARGVAL-1', 'Top icon', '#tag', '#tag']
Anarchism: WARNING: invalid attribute format 'This is a good article. Click here for more information.]]' missing name at ['Anarchism', 'good article', 'Main other', 'ARGVAL-1', 'Top icon', '#tag', '#tag']
[...]

Issue #90, opened 2023-08-25
How to get text from from templates?


rusg77:
Hi, thanks for the project! 

I'm trying to extract text from wiki dumps. Page example: https://en.wikipedia.org/wiki/Free_neutron_decay
I dowloaded the page and it's templates via https://en.wikipedia.org/wiki/Special:Export
The page contains the following template - `{{val|879.6|0.8|u=[[second|s]]}}` which I'd like to be converted to `879.6±0.8 s` text.

Code (the latest in repo):

```
    def test_simple_page(self):

        def page_handler(page: Page, wtp: Wtp | None = None) -> Any:
            wtp.start_page(page.title)
            node = wtp.parse(page.body, pre_expand=True)
            value = wtp.node_to_wikitext(node)
            print(value)

        wtp = Wtp(db_path=Path('../db/mydb'))
        process_dump(
            wtp,
            "../Wikipedia-20230825082710.xml.bz2",
            {0, 10, 110, 828},  # namespace id
            save_pages_path=Path('../pages')
        )

        print("Num pages:", wtp.saved_page_nums())

        for _ in map(
                partial(page_handler, wtp=wtp), wtp.get_all_pages([0])
        ):
            pass
```
Output:
```
Num pages: 86
.... <strong class="error">Template:val</strong>...

```

If I set `pre_expand` to `False`, then:
```
Num pages: 86
.... {{val|879.6|0.8|u=[[second|s]]}}...
```

Probably it's smth simple, but I can't find a solution. Can you please help?

Github doesn't allowed to upload xml/bz2 files, so I uploaded my xml on Dropbox: [link](https://www.dropbox.com/scl/fi/ftbygha0eovyz5xl3rc6y/Wikipedia-20230825082710.xml.bz2?dl=0&rlkey=kd43093wgdd2sq08rvras32rf)

kristian-clausal:
I'll try to take a look at this next week. Please also give more context around test_simple_page(self) (is this from a test_ file?), and whatever output you get. Do the `....` mean you actually get output around the templates, for example, etc.

rusg77:
Thanks @kristian-clausal !

> Please also give more context around test_simple_page(self) (is this from a test_ file?)

I just tried to run my code as test, but it doesn't matter. So the code actually:
```
from functools import partial
from pathlib import Path
from typing import Any

from wikitextprocessor import Wtp, Page
from wikitextprocessor.dumpparser import process_dump


[...]

PR #271, opened 2024-04-18
Run `mw.loadData()` in cloned environment

xxyzz:
As the Scribunto code and the GH issues(#90, #258) shows, `mw.loadData()` runs in a cloned environment so the caller env doesn't affect the code runs inside `loadData`.

I finally sort of understand this clone env problem...


kristian-clausal:
Wonderful!!

xxyzz:
I have tested the code on es edition dump file and there are no new Lua errors.

PR #273, opened 2024-04-22
Create `Logger` object and set `Logging.DEBUG` level

xxyzz:
Our previous code use the root logger configed in wiktextract code, but now wiktextract doesn't use root logger thus all info level logs are not printed.

### DIFF
--- a/src/wikitextprocessor/core.py
+++ b/src/wikitextprocessor/core.py
@@ -42,6 +42,7 @@ from .common import (
     add_newline_to_expansion,
     nowiki_quote,
 )
+from .logging_utils import logger
 from .luaexec import call_lua_sandbox
 from .node_expand import NodeHandlerFnCallable, to_html, to_text, to_wikitext
 from .parser import (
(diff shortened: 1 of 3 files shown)


kristian-clausal:
I missed this, but I wonder if we need to set a logging level for the wikitextprocessor logger from inside wiktextract? Does INFO get a higher precedence than DEBUG, so INFO and WARNING will get through?

xxyzz:
info is higher than debug: https://docs.python.org/3/library/logging.html#logging-levels so set level to debug enables info, default level is probably warning.

I think we're current use logging as `printf` and our default level is debug, so don't combine two packages' logger is simpler. Maybe we could consider how to use the logging library later if we don't want two separate log methods...

kristian-clausal:
I meant that we might have to modify wikitextprocessor's logger's level, like we do with other loggers. It is correct to have two different loggers for wiktextract and wikitextprocessor, because they're two packages.

xxyzz:
I'll pass the `--verbose` bool command option to `Wtp.__init__`, and set the logger debug level if it's `False` so we'll have the save level in two packages.

xxyzz:
I added an extra `not` and wiktextract uses the `--quite` option...

PR #277, opened 2024-04-25
Remove the limit of unnamed template argument number

xxyzz:
0 and above 1000 are allowed in MediaWiki, for example, fr edition's "cardinaux/vi" template:
https://fr.wiktionary.org/wiki/Modèle:cardinaux/vi

### DIFF
--- a/src/wikitextprocessor/core.py
+++ b/src/wikitextprocessor/core.py
@@ -1602,13 +1602,6 @@ class Wtp:
                             k, arg = m2.groups()
                             if k.isdigit():
                                 k = int(k)
-                                if k < 1 or k > 1000:
-                                    self.debug(
-                                        "invalid argument number {} "
-                                        "for template {!r}".format(k, name),
-                                        sortid="core/1211",
-                                    )
-                                    k = 1000
                             else:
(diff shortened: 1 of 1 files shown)


kristian-clausal (changes requested):
Those are some big numbers. This seems obviously correct to me, and if the numbers go that high I bet the limit is actually something like `int`, in which case it doesn't really make sense to check upper limits.

However, we might need to check for lower, so there should still be a `k < 1` test. In that case you might as well leave the deleted block and replace the magic number `1000` with MAX_TEMPLATE_KEY_VALUE_INT or something similar, and make the constant the size of an unsigned(?) int or whatever makes most sense... If there is anything that makes the most sense. PHP doesn't seem to have unsigned ints, but it does have different sizes for 32 and 64 bit... Yeah, I guess at this point it doesn't make sense to check max, just `k < 1`.

xxyzz:
0 is also allowed and `str.isdigit()` returns `False` for negative string number so I think we don't need more checks.

kristian-clausal (approved):

kristian-clausal:
Then everything seems correct.

PR #281, opened 2024-05-07
Don't add too many template args debug message for empty args

xxyzz:
It's not uncommon in non-en editions to have `{{{x||}}}` in templates.


kristian-clausal (approved):
This seems perfectly fine. :+1:

PR #285, opened 2024-05-22
Change external links `[...]` regex

kristian-clausal:
`[test]]` is not a valid link, so prevent that with negative lookahead. This caused problems in cases like:

  {{quote-book|en|author=[[w:Theodore Beza|Theodore de Beza]]
  |tlr=[[w:Robert Fills|R[obert] F[ills&#93;]]......

with the open `[` bracket messing with the last `]]` token, causing the later LINKS_RE pattern call to basically take ages.

### DIFF
--- a/src/wikitextprocessor/core.py
+++ b/src/wikitextprocessor/core.py
@@ -194,7 +194,8 @@ LINKS = (
 LINKS_RE = re.compile(LINKS)
 
 # Encode external links: [something]
-EXTERNAL_LINKS = r"\[([^][{}<>|\n]+)\]"
+# [test]]+ breaks in sandbox, so prevent with negative lookahead
+EXTERNAL_LINKS = r"\[([^][{}<>|\n]+)\](?!\])"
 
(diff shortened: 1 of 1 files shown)


xxyzz:
More information: the original wikitext is `[[w:Robert Fills|R[obert] F[ills&#93;]]` in page [earnestlier](https://en.wiktionary.org/wiki/earnestlier) and many author links in example templates are written in this format.

Don't know why they're not using `[[w:Robert Fills|Robert Fills]]` but I guess the use of `&#93;]]` is because(maybe) a MediaWiki bug: `[[w:Robert Fills|R[obert] F[ills]]]` can't render correctly in MediaWiki.

kristian-clausal:
I think `[[w: RF|...F[ills]]]` would render as `Robert Fills]`. The `]]` is eaten by the link, and the last `]` is orphaned, because the `[brackets]` aren't parsed as external links (no url). This is expected behavior, it was ours that was wrong because our external link regex ate `[ills&#93;]`, breaking up the `]]` token.

PR #287, opened 2024-06-06
Remove upper case substitution modifiers

xxyzz:
an upper case "SAFESUBST:" is used in https://en.wiktionary.org/wiki/Template:ordinal

Issue https://github.com/tatuylonen/wiktextract/issues/663


kristian-clausal:
Thanks, Tatu's been meaning to take a look at subst: and safesubst: in-depth for some time now, but he's been busy with other things. I know I wasn't able to fix the issues with them.

PR #290, opened 2024-06-19
Implement `#rel2abs` parser function

xxyzz:
Used in ru edition's "Шаблон:родств-блок" template. Doesn't affect extracted text but remove many unimplemented error messages.

I only consider the common use cases, MediaWiki's code are complex and we don't need to handle all the edge cases at the moment.


kristian-clausal:
This looks good to me :+1:

PR #291, opened 2024-06-21
Only parse `----` as horizontal rule if it's at the start of line

xxyzz:
https://en.wikipedia.org/wiki/Help:Wikitext#Horizontal_rule

GitHub issue tatuylonen/wiktextract#536


kristian-clausal:
Good catch!

kristian-clausal (commented):
Looks good to me. This is such a no-brainer that I can't think of how it could be wrong.

PR #297, opened 2024-08-07
Call `Wtp.preprocess_text()` again after calling `Wtp.expand()` in `Wtp.parse()`

xxyzz:
Template could contain `<nowiki/>` tag and needs to be converted, otherwise the `<span>` tag is parsed as text but not `HTMLNode`.

Find the bug in pl edition "etym" template.

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -2932,6 +2932,21 @@ def foo(x):
             {1: "en", 2: "—hole", 3: "----hole"},
         )
 
+    def test_nowiki_in_html_attr_value(self):
+        # https://pl.wiktionary.org/wiki/Szablon:skrót/szkielet
+        # used in etymology template https://pl.wiktionary.org/wiki/Szablon:etym
+        self.ctx.start_page("pies")
+        self.ctx.add_page(
+            "Template:skrót/szkielet",
+            10,
+            '<span class="short-container<nowiki/> ">text</span>',
+        )
+        root = self.ctx.parse("{{skrót/szkielet}}", expand_all=True)
+        self.assertEqual(len(root.children), 1)
+        span_node = root.children[0]
+        self.assertIsInstance(span_node, HTMLNode)
+        self.assertEqual(span_node.tag, "span")
+
 
 # XXX implement <nowiki/> marking for links, templates
 #  - https://en.wikipedia.org/wiki/Help:Wikitext#Nowiki
--- a/src/wikitextprocessor/core.py
+++ b/src/wikitextprocessor/core.py
@@ -1985,6 +1985,7 @@ class Wtp:
             text = self.expand(
                 text, template_fn=template_fn, post_template_fn=post_template_fn
             )
+            text = self.preprocess_text(text)
             # print(f"PARSE EXPAND ALL: {text=!r}")
         elif pre_expand or additional_expand:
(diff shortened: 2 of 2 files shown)


xxyzz:
wiktextract tests also passed.

tests/test_parser.py
```diff
@@ -2932,6 +2932,21 @@ def test_horizontal_rule_in_template_arg(self):
...
         )
 
+    def test_nowiki_in_html_attr_value(self):
+        # https://pl.wiktionary.org/wiki/Szablon:skrót/szkielet
+        # used in etymology template https://pl.wiktionary.org/wiki/Szablon:etym
+        self.ctx.start_page("pies")
+        self.ctx.add_page(
+            "Template:skrót/szkielet",
+            10,
+            '<span class="short-container<nowiki/> ">text</span>',
```
kristian-clausal, line 2942: I wonder why the nowiki is used here.

kristian-clausal:
This seems fine. I wonder if we could make preprocess_text more efficient by checking `if "nowiki" in in text` for the two re.sub calls, but that might actually make it less efficient if `nowiki` is almost always present. AFAICT, _nowiki_sub_fn() doesn't need to be a nested function here, and if it's recreated every time preprocess_text is called that seems inefficient, so we might want to move it out. Do you know about this kind of optimization in Python? I actually thought the nested functions were "cached"...

I think Tatu might think the nested function isn't recreated due to this background being in compiled languages.

We might see some performance improvements if we move nested functions that don't rely on closure-related stuff outside of their main function, and change some of the closure-using functions into ones that use normal parameters instead. It would make the code less readable.

That's outside the scope of this.

xxyzz:
I think we probably don't need to worry about optimize nested function for speed, unless profile result shows it's too slow.

But now I find with this pr change, `clean_node()` still returns broken text after the `nowiki` tag. `clean_node` uses `Wtp.node_to_html()`, and `node_to_html()` returns texts contain `<nowiki/>`, I guess this is correct? So the problem is at the `clean_value()` function?

xxyzz:
regex at here: https://github.com/tatuylonen/wiktextract/blob/4703eb899fc7395acdefabbee2bde249f9e7cf96/src/wiktextract/clean.py#L1443-L1447

can't match `<span class="short-container<nowiki/> other-class">text</span>` correctly, damn...

kristian-clausal:
I don't think there's any harm in removing `<nowiki/>` in clean_value, somewhere at the start of the function. I think the assumption is that `<nowiki/>` has already been handled before this, but this case slips through. But if we can do this properly, it would be the best.

xxyzz:
Do you know where `<nowiki/>` is removed? I think I have seen it removed at somewhere but can't find it anymore. But I guess it's not wrong if `Wtp.node_to_html()` returns text contains `<nowiki/>`, maybe even expected?

kristian-clausal:
Sorry for the late reply, I was out for a moment. It makes sense that `node_to_html()` returns `<nowiki/>` intact, yeah.

I think nowiki is removed in the process of parsing in the sense that when nowiki stuff is converted into a parse tree, it's then later ignored and not output. The process of creating the parse tree removes the nowiki as it is processed, and the parser shifts to a nowiki-state that outputs raw text.

PR #302, opened 2024-09-05
Don't add empty argument to parser functions in `to_wikitext()`

xxyzz:
wiktextract code `clean_node()` calls `Wtp.node_to_html()` and finally to `to_wikitext()`.

This function converts `{{PAGENAME}}` to `{{PAGENAME:}}`, and the second form expands to empty string but not page title.

Find this error when fixing ja edition exceptions.

All wiktextract tests passed.


kristian-clausal:
Good catch!

PR #304, opened 2024-09-11
Only parse external link as text if `<nowiki/>` directly after `[`

xxyzz:
Put `<nowiki/>` at any other place expands to normal external url link.

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -2947,6 +2947,21 @@ def foo(x):
         self.assertIsInstance(span_node, HTMLNode)
         self.assertEqual(span_node.tag, "span")
 
+    def test_nowiki_tag_in_external_link(self):
+        # https://zh.wiktionary.org/wiki/Template:RQ:Qur'an
+        self.ctx.start_page("محمد")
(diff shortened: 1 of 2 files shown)


tests/test_parser.py
```diff
@@ -2947,6 +2947,21 @@ def test_nowiki_in_html_attr_value(self):
...
+        self.assertIsInstance(url_node, WikiNode)
+        self.assertEqual(url_node.kind, NodeKind.URL)
+        self.assertEqual(
+            url_node.largs, [["https://quran.com/3/144"], ["3<nowiki />:144"]]
+        )
+
+        root = self.ctx.parse("[ <nowiki/> https://quran.com/3/144 3:144]")
+        text_node = root.children[0]
+        self.assertEqual(text_node, "[ <nowiki /> ")
+
```
kristian-clausal, line 2964: I have confirmed that this is correct behavior. In addition to this, doing `[https://goo<nowiki/>gle.com test]` in the sandbox results in `[https://goo^gle.com test]` as text with `https://goo` parsed as an URL, but it seems extremely unlikely that this kind of thing is done on purpose.

kristian-clausal (approved):
This is correct, but I've mentioned an edge-case that might be relevant.

PR #309, opened 2024-09-19
Return empty string for `{{ns:0}}` and `{{ns:}}`

xxyzz:
This is required in ru edition's language title template "--lang--" to create language category link.


xxyzz:
Previous code returns "Main" from these namespace JSON files. I forget why it uses "Main" for the namespace name, maybe it's always "Main" in our code.

PR #311, opened 2024-09-26
Don't let section start token break unclosed parent node

### LINKED ISSUE #310: Section node in template parameter shouldn't break parsing template node
There are some `TEMPLATE not properly closed` debug messages in fr edition: https://kaikki.org/frwiktionary/errors/subpage6/details-TEMPLATE-not-properly-closed.html

Here is part of wikitext from page https://fr.wiktionary.org/wiki/Conjugaison:français/bayer

```
{{Onglets conjugaison
| onglet1  =/ba.je/ avec « y » (je baye)
| contenu1 ={{fr-conj-1-yer-ye|ba|ba}}
| onglet2  =/ba.je/ avec « i » (je baie)
| contenu2 ={{fr-conj-1-yer-ie|ba|b|a|ɛ|ɛ}}
=== {{S|références}} ===
* Frères Bescherelle (sans spécifier la prononciation des finales en ''-aie'')
| onglet3  =/be.je/ avec « y » (je baye)
}}
```

I think the code at here: https://github.com/tatuylonen/wikitextprocessor/blob/66545a619dac146e47ea81e9cc13872b890ad97e/src/wikitextprocessor/parser.py#L997-L1003

shouldn't remove nodes in parse stack if the node needs a close token. In other words, section heading node start token shouldn't break parsing other nodes if it's inside a node.

Maybe the code should be:

```python
    while any(x.kind in KIND_TO_LEVEL for x in ctx.parser_stack):
        node = ctx.parser_stack[-1]
        if KIND_TO_LEVEL.get(node.kind, 99) < level:
            break
        if node.kind in MUST_CLOSE_KINDS:
            break
        _parser_pop(ctx, True)
```

I have tested the code and all tests in both packages passed.

xxyzz:
it shouldn't pop nodes in parser stack if it's inside an unclosed node like a template

fixes #310

### DIFF
--- a/src/wikitextprocessor/parser.py
+++ b/src/wikitextprocessor/parser.py
@@ -259,27 +259,27 @@ LEVEL_KIND_FLAGS = (
 )
 
 # Node types that have arguments separated by the vertical bar (|)
-HAVE_ARGS_KINDS: tuple[NodeKind, ...] = (
-    NodeKind.LINK,
-    NodeKind.TEMPLATE,
-    NodeKind.TEMPLATE_ARG,
-    NodeKind.PARSER_FN,
-    NodeKind.URL,
+HAVE_ARGS_KIND_FLAGS = (
+    NodeKind.LINK
+    | NodeKind.TEMPLATE
+    | NodeKind.TEMPLATE_ARG
+    | NodeKind.PARSER_FN
+    | NodeKind.URL
 )
 
 
 # Node kinds that generate an error if they have not been properly closed.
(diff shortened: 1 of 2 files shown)


kristian-clausal (approved):
If the test at the end seems to pass, the problem in the last post in the issue (about `contenu3` being the child of the level 3 node?) seems to be fixed? Otherwise, all these changes seem correct. Sometimes there is no correct way to do things, and I suspect the way Wikitext works might be one of them.

PR #315, opened 2024-10-01
Handle includeonly elements

### LINKED ISSUE #314: Newlines inside `includeonly` are expanded on our side, but not in wikitext.
For example:
[Template:ti verb](https://simple.wiktionary.org/wiki/Template:ti_verb)
```
(''[[transitive|<span style="color:green">transitive</span>]] & [[intransitive|<span style="color:green">intransitive</span>]]'')<includeonly>[[Category:Transitive verbs]]
[[Category:Intransitive verbs]]</includeonly><noinclude>
{{documentation}}
<!-- Categories, interwikis and TemplateData goes to the /doc subpage -->
</noinclude>
```

is used on one page [excrete](https://simple.wiktionary.org/wiki/excrete). The `includeonly` element has a newline, which is rendered on our side as `'glosses': ['(transitive & intransitive)\n Foooo.']`. The `Template:biology` on the same page doesn't get the newline, because there is no newline inside the `includeonly`.

I mixed up `onlyinclude` and `includeonly` for a while and it took me a while to understand which is which... `onlyinclude` is text that is the only thing you want to output when the template is expanded (anything outside of it is discarded). `includeonly`, which is what is at issue here, is a piece of text like a Category link that you don't want to appear on the *template's own display page*. That is, you don't want `Template:ti verb` to be appear in the `Transitive verbs` category, so `includeonly` will only let the category link be rendered when the template is being expanded on some other page.

However, I don't understand why the newline disappears in this case. Either this is so common in wikitext that they just went ahead and removed all newlines in onlyinclude, or something else weird.

**Has a newline after "transitive)" on Wiktionary**
```
(''[[transitive|<span style="color:green">transitive</span>]] & [[intransitive|<span style="color:green">intransitive</span>]]'')<includeonly>


[[Category:Transitive verbs]]

[[Category:Intransitive verbs]]
</includeonly><noinclude>
{{documentation}}
<!-- Categories, interwikis and TemplateData goes to the /doc subpage -->
</noinclude>
```

**Does not have a newline after "transitive)"Wiktionary**
```
(''[[transitive|<span style="color:green">transitive</span>]] & [[intransitive|<span style="color:green">intransitive</span>]]'')<includeonly>


[[Category:Transitive verbs]]

[[Category:Intransitive verbs]]</includeonly><noinclude>
{{documentation}}
<!-- Categories, interwikis and TemplateData goes to the /doc subpage -->
</noinclude>
```

kristian-clausal:
Fixes #314

Apparently <includeonly> (which renders text only when used as a template (transcluded), not when showing the template's own page so that these pages don't get
stuff like wrong categories) has some weird whitespace trimming rules.

1. If there's a newline before the end tag, return everything and add a space (because this causes a PRE block to appear??)
2. Otherwise, if there's only links or other things that render as whitespace strip all the whitespace.
3. If there's other text, only strip away the whitespace that is after that text, ignoring category links.

### DIFF
--- a/src/wikitextprocessor/core.py
+++ b/src/wikitextprocessor/core.py
@@ -238,6 +238,49 @@ ALL_BRACKETS_RE = re.compile(
     + r")"
 )
 
+INCLUDEONLY_RE = re.compile(r"(?is)<includeonly\s*>(.*?)(\n?)</includeonly\s*>")
+
+CATEGORY_LINK_RE = re.compile(r"(?i)(\[\[ *Category *:.*?\]\])")
+
+
+def includeonly_re_fn(m: re.Match) -> str:
+    """Try to emulate how Wikitext inside includeonly is processed: if the
+    end tag is preceded by a newline, just return everything almost as is; if
+    there is non-Category link texts otherwise, then keep whitespace from the
+    start until the end of the "visible" text. Otherwise remove whitespace and
+    newlines."""
+    assert m is not None
+    if m.group(1) == "":
+        return ""
+    if m.group(2) != "":
+        # There was a newline before the </includeonly>, which seems to
+        # trigger this, and a weird space that causes <pre>-blocks.
+        return m.group(1) + "\n "
+    rev_new_parts: list[str] = []
+
+    # split and reverse: ["...", "[[Category:...]]", "...", cat, ...]
+    rev_old_parts: list[str] = list(
+        reversed(CATEGORY_LINK_RE.split(m.group(1)))
+    )
+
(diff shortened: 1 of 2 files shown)


kristian-clausal:
This is such a hack... If this goes through, .db files might need updating for it to take effect, because this affects dump parsing / page saving. I do not want to add parsing code into the parser to handle this stuff... D: One and done during .db generation seems appropriate.

kristian-clausal:
As discussed in #314, this was not an issue with `includeonly`, but apparently how whitespace or newlines are trimmed around Category links.

PR #319, opened 2024-10-03
Only parse section node if the token is at the beginning of a line 

xxyzz:
Fix zh edition exception: https://kaikki.org/zhwiktionary/errors/details-----EXCEPTION-while-parsing-page-----i-YTyJLi-P.html


kristian-clausal (approved):
Weird how this hasn't turned up before, but yeah, the tokenizer just looks at raw "===="s, there was no checking for start of line... This all seems correct!

xxyzz:
I also thought this should be fixed a long time ago...

PR #320, opened 2024-10-07
Implement "int" parser function

xxyzz:
It's used in Dutch Wiktionary's "ISOdate" Lua module, return language code for `{{int:lang}}` in a Wiktionary project.


kristian-clausal:
Oh it's "interface"... At first I thought integer, then international, but it's about namespaces. I wonder if it's used anywhere besides this `lang` use.

xxyzz:
It's used for retrieving some values defined for a MediaWiki server, I guess it's created by server admins. I have seen this function used in other edition Lua code.

PR #321, opened 2024-10-07
Allow "div" tag inside "ref" tag

xxyzz:
Dutch Wiktionary's example template expands this kind of HTML tags, previous code move the "div" tag outside the "ref" tag and break the "ref" tag, make it unclosed.

I don't really know what's the difference between "flow" and "phrasing", but this works.

### DIFF
--- a/src/wikitextprocessor/wikihtml.py
+++ b/src/wikitextprocessor/wikihtml.py
@@ -59,7 +59,7 @@ ALLOWED_HTML_TAGS: dict[str, HTMLTagData] = {
     },
     "del": {"parents": ["phrasing"], "content": ["phrasing"]},
     "dfn": {"parents": ["phrasing"], "content": ["phrasing"]},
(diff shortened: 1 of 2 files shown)


kristian-clausal:
"phrasing" and "flow" are HTML element categories, they're not something we've defined here. https://developer.mozilla.org/en-US/docs/Web/HTML/Content_categories

'ref' is messing with it because it's a Wikitext-specific not-HTML tag, which only looks like an HTML tag. We might need to do special handling for "ref", or maybe invent a "wikitext" category for it, or make "ref" "flow"... Except "ref" can appear where "flow" doesn't.

kristian-clausal:
Yeah, I asked Tatu and changing the category of `div` will lead to bugs. The best way we could think of doing this is to create a new "category", something like `wikitag` and let it act as a parent and a child to anything and everything. I will quickly revert this PR and we can take a look tomorrow.

PR #323, opened 2024-10-07
[nl] Change 'ref' HTML 'category' to '*'

kristian-clausal:
Continued from PR #321

There are a few wikitext-specific "HTML-looking" tags like `includeonly` that have "*" as their category, allowing them to be used anywhere in the text.

Co-authored by: xxyzz <gitpull@protonmail.com>


xxyzz:
I should wait a bit longer... I didn't realize "ref" is not a HTML tag and I also didn't know "*" could be used. Thanks for the correction! I have merged your commit to the main branch, I also dropped my previous commit and added the MDN link to comments.

PR #326, opened 2024-10-15
Overload LevelNode.find_content with TEMPLATE variant

kristian-clausal:


xxyzz:
I'm not quite understand why we need this kind of overload, mypy couldn't be this useless...

kristian-clausal:
This is actually logical, because Mypy can't make the connection (because there is technically none) between NodeKind.TEMPLATE and TemplateKind. That is pretty high level logic. Check out de/page.py/129

```python
def process_pos_section(
    wxr: WiktextractContext,
    page_data: list[WordEntry],
    base_data: WordEntry,
    level_node: LevelNode,
) -> None:
    pos_data_list = []
    for template_node in level_node.find_content(NodeKind.TEMPLATE):
        if template_node.template_name == "Wortart":
```
Mypy will complain that `WikiNode` doesn't have `.template_name` (which is true), and it can't know that `.find_content()` always returns `TemplateNode` when given `NodeKind.TEMPLATE` because that is not a given. In this case, it's asking a bit too much from mypy.

xxyzz:
Ah, the return type. Maybe the wikitextprocessor API could be better designed for this kind of problem...

PR #328, opened 2024-10-18
find_html: allow list of tags to find.

kristian-clausal:
To make find_html work the same was as find_child, allow `target_tag` (now `target_tags`) to be a list of strings instead of a single string, and return
all HTML element children with one of those tags.


kristian-clausal:
Ah, what an embarrassing mistake... Thank you, --amend and push --force for hiding it from the world.

xxyzz (approved):

PR #333, opened 2024-11-13
Detect template loop formed in Lua code

xxyzz:
Fix tatuylonen/wiktextract#894

`test_template16` is removed because it's not a valid test, it has infinite loop.

I can't reproduce the loop in test, but it has the same repeated expand stack as the error message. The test stops at "too deep recursion" error without the change but it fixes the loop in page "今生".

### DIFF
--- a/src/wikitextprocessor/core.py
+++ b/src/wikitextprocessor/core.py
@@ -1410,7 +1410,18 @@ class Wtp:
                         continue
 
                     # Construct and expand template arguments
-                    self.expand_stack.append(name)
+                    self.expand_stack.append("Template:" + name)
+                    if detect_expand_template_loop(self.expand_stack):
+                        parts.append(
+                            '<strong class="error">Template loop detected:'
+                            f"[[:Template:{name}]]</strong>"
+                        )
+                        self.expand_stack.pop()
+                        self.warning(
+                            f"Template loop detected: {name}",
+                            sortid="core/1422",
+                        )
+                        continue
                     ht: TemplateArgs = {}
                     num = 1
                     for i in range(1, len(args)):
@@ -1450,7 +1461,9 @@ class Wtp:
                     t: Optional[str] = None
                     # print("EXPANDING TEMPLATE: {} {}".format(name, ht))
                     if template_fn is not None:
+                        self.expand_stack.append("TEMPLATE_FN")
                         t = template_fn(urllib.parse.unquote(name), ht)
+                        self.expand_stack.pop()
                         # print("TEMPLATE_FN {}: {} {} -> {}"
                         #      .format(template_fn, name, ht, repr(t)))
                     if t is None:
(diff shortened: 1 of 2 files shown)


src/wikitextprocessor/core.py
```diff
@@ -1909,3 +1918,14 @@ def create_strip_marker(self, node: str, content: str) -> str:
                 self.strip_marker_cache["preprocess"] += 1
                 self.strip_marker_cache[content] = num
         return f"""\x7f'"`UNIQ--{node}-{num_str}-QINU`"'\x7f"""
+
+
+def detect_expand_template_loop(stack: list[str]) -> bool:
+    stack_len = len(stack)
```
kristian-clausal, line 1924: Do you think we could add a couple of guard clauses here? Like `if stack[-1] not in stack[:-1]: return False`? We could also return True when stack_len is small, and only start checking when it grows too big.

src/wikitextprocessor/core.py
```diff
@@ -1410,7 +1410,14 @@ def expander(arg: str) -> str:
                         continue
 
                     # Construct and expand template arguments
-                    self.expand_stack.append(name)
+                    self.expand_stack.append("Template:" + name)
+                    if detect_expand_template_loop(self.expand_stack):
+                        parts.append(
```
kristian-clausal, line 1415: Should we have a `wxr.wtp.error` or `wxr.wtp.warning` here? We are handling the issue, and it's caused by bad code on Wiktionary's side, so not necessarily.

kristian-clausal (approved):

kristian-clausal:
I had only minor comments that I'm not sure are valid, otherwise this looks good. I'm worried how performant the loop checking is, but that's not something we can avoid easily.

xxyzz:
Add `mw.logObject(template:get_arguments())` at line 93 in https://en.wiktionary.org/wiki/Module:zh-translit and preview page "今生", the Lua log will show loop detected for template "m".

kristian-clausal:
Do you think they know about this on the Wiktionary side of things?

xxyzz:
It's not a problem on Wiktionary because the parser could stop the loop, I think maybe Lua module editors don't need to worry about it.

src/wikitextprocessor/core.py
xxyzz: Short list also could have loop and check it is fast. I'll add the "if" check.

src/wikitextprocessor/core.py
xxyzz: I'll add a warning here.

src/wikitextprocessor/core.py
xxyzz: Maybe there is a better algorithm for this function, I'm also not sure how this would impact performance.

kristian-clausal: We can merge now that it's working, see how much performance is affected and try to figure out something later.

xxyzz:
There is a discussion about the bug on Wiktionary: https://en.wiktionary.org/wiki/User_talk:Theknightwho#Pages_with_template_loops

PR #335, opened 2024-11-14
Don't treat nested template in template parameters as expansion loop

xxyzz:
also find this false positive bug in en edition's [quote-song](https://en.wiktionary.org/wiki/Template:quote-song) template, it has some nested "ifmatch" templates.

### DIFF
--- a/src/wikitextprocessor/core.py
+++ b/src/wikitextprocessor/core.py
@@ -1925,13 +1925,17 @@ class Wtp:
 
 
 def detect_expand_template_loop(stack: list[str]) -> bool:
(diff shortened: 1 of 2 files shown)


kristian-clausal (approved):
This makes sense, nothing much to add.

kristian-clausal:
Merged this because I finally noticed that kaikki regen was stuck...

xxyzz:
Hope this will fix the website update bug, but I'm not sure if they are related...

kristian-clausal:
It was the loop, I just wanted this to be in there for completion's sake.

PR #337, opened 2024-11-15
Load Lua module namesapce page not standard library for `#invoke`

xxyzz:
Fix Lua error in [Template:R:gem:HGE](https://en.wiktionary.org/wiki/Template:R:gem:HGE), Lua's math library doesn't have `sum` function, should load "Module:math".


kristian-clausal (approved):
This seems correct to me :+1:

Issue #336, opened 2024-11-14
Line with only spaces parsed as PREFORMAT


timendum:
Hi,
I'm still working on processing the italian wiki, but I found an issue on the parser.

Example page: https://it.wiktionary.org/wiki/gioved%C3%AC

The source is:
```
; gio | ve | dì
\s\t  	
{{-pron-}}\t
{{IPA|/dʒoveˈdi*/}} (causa il raddoppiamento della consonante successiva)
* '''''giovedì''' prossimo'' {{IPA|/dʒoveˌdi pˈprɔssimo/}}
```
_(I transformed spaces and tabs in \s and \t, for clarity)_

Here the "\s\t" row is parsed as "PREFORMATTED", but WikiMedia preformat only if there is text in the line.

It's possible to fix?

xxyzz:
I also encounter a similar problem in Dutch Wiktionary, a space at line start parsed as preformat node and breaks other nodes:

```
::''list<ref> NRC Bart Hinke
 24 augustus 2015 </ref>'
```

form page [doodgeboren](https://nl.wiktionary.org/wiki/doodgeboren)

xxyzz:
I think preformatted node should be popped from `Wtp.parser_stack` if the next line is not start with space at here https://github.com/tatuylonen/wikitextprocessor/blob/50313cf9c277d21f89035b3da181b1b7c2fa131a/src/wikitextprocessor/parser.py#L2354-L2356

but some tests fail and need to be updated.

kristian-clausal:
These two seem like separate issue.

1. I checked in the sandbox and indeed `\s\t` doesn't generate `<pre>` in the HTML source, so we shouldn't either; probably need to do a check with strip() to see if there's any text on the line.
2. If we keep the PRE node open on subsequent lines after a newline when there's no space, that's wrong and should be fixed.

kristian-clausal:
I've committed a fix to the ref issue, and will merge it next week because it's Friday afternoon.

kristian-clausal:
@timendum sorry for getting to this this late, I completely forgot this issue was still open. How is the situation right now? If you find anything that seems still broken, just post here and I'll reopen this issue.

timendum:
No issue so far, thanks!

PR #338, opened 2024-11-15
Ignore contents of whitespace-only lines

kristian-clausal:
Issue #336

A line with `\s\t` should not trigger a PREFORMATTED block. The easiest way to handle this is to just
ignore the contents of whole lines when they're just whitespace.

The regex split that splits the lines on "(\n+)"
means that no 'line' contents has newlines characters, so we need to take a look at any tokenizing regex that has `\n` in it; those newlines don't do anything.

The newlines are tokenized because re.split, when given a capture group, will alternate between the splittéd and splitter texts: ["text", "\n", "text", "\n\n"].


kristian-clausal:
This only fixes the issue in the first post with tabs.

PR #339, opened 2024-11-15
`<ref>` should ignore whitespace preformatted lines

kristian-clausal:
Fixes second part of #336

When we would normally be pushing a layer of NodeKind.PREFORMATTED, check if the stack has any HTML-elements with the `<ref>` tag and if that is the case don't push PREFORMATTED.

If other HTML-like tags need this in the future (onlyinclude? includeonly??) then this is the place where to add those checks.

Looking through the stack could be better. Filter? Any? Oh yeah, duh.

### DIFF
--- a/src/wikitextprocessor/parser.py
+++ b/src/wikitextprocessor/parser.py
@@ -967,7 +967,14 @@ def text_fn(ctx: "Wtp", token: str) -> None:
             ):
                 return
             # print(f"{token=}")
-            if node.kind != NodeKind.PREFORMATTED and not ctx.pre_parse:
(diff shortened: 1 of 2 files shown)


xxyzz (approved):

xxyzz:
Thanks for the fix! And I also add the "p" tag to the "if" check line.

kristian-clausal:
Thanks for finishing this for me while I was away, lol, I just figured out what was happening; in my branch, I had a commit that removed the for loop and replaced it with `any()`, and I was confused why I was getting merge conflicts.

PR #345, opened 2024-11-28
Add test for HTML comment inside list item parsing

kristian-clausal:


xxyzz:
Is this the same test added in #344?

kristian-clausal:
... I FORGOT TO PULL WIKITEXTPROCESSOR. 🤦 I just remembered "Oh yeah, I meant to add that test to this issue two weeks ago", checked the code... And it was old code without #344. I will revert this.

PR #346, opened 2024-11-29
Fix two string index out of range exceptions

xxyzz:


kristian-clausal:
https://fr.wiktionary.org/wiki/Mod%C3%A8le:ga-tab-conj- doesn't seem to exist; maybe it was deleted, or `Modèle:ga-tab-conj-` with a dangling `-` at the end is an incomplete generated template name?

xxyzz:
Should be https://fr.wiktionary.org/wiki/Modèle:ga-tab-conj-1, I missed the number somehow.

PR #351, opened 2024-12-11
Expand template changes for Italian Wiktionary

xxyzz:
Minor changes for it edition's section title templates, shouldn't affect en code.


kristian-clausal (approved):
If this is what's needed for Italian wiktionary to work... But it hurts. More kludges..!

xxyzz:
Um, I think the changes are not really ugly kludge, it's alright...

Issue #352, opened 2024-12-17
Parsing typoed `={-hr-}==` has weird behavior


kristian-clausal:
Example taken from Greek Wiktionary article peta, where we have the (typoed) code

```
=={{-id-}}==

==={{ουσιαστικό|id}}===
{{τ|id|{{PAGENAME}}}}
* [[χάρτης]]

===={{συνώνυμα}}====
* [[map]]


----

={{-hr-}}==

==={{ουσιαστικό|hr}}===
{{τ|hr|{{PAGENAME}}}} {{θ}}
* [[φτέρνα]]


----

=={{-sl-}}==
...
```

EDIT: Added more context here, horizontal lines have some things going with parsing levels.

The parse tree should look similar to another header tree:

```
<LEVEL2([<TEMPLATE(['-id-']){} >]){}
   '\n\n',
   <LEVEL3([<TEMPLATE(['ουσιαστικό'], ['id']){} >]){}
      '\n', <TEMPLATE(['τ'], ['id'], [<PARSER_FN(['PAGENAME']){} >]){} > ...>>
```

but instead it looks like:

```
<LEVEL1(){}
   <TEMPLATE(['-hr-']){} >,
   '==\n\n',
   <LEVEL3([<TEMPLATE(['ουσιαστικό'], ['hr']){} >]){} 
      '\n', ...>>
```

The `{{-hr-}}` template that should be in `.largs` (like the `{{-id-}}` template for the Indonesian heading) is instead a direct child of the LEVEL1. It should be inside the parentheses, not after the `{}`. I would also expect only one `=`, but both of them are there.

Posting this as a todo for myself, it's getting late.

xxyzz:
I have seen similar broken wikitext and I always fix them on wiktionary, they are rare. We could find these pages in the "can't find subtitle start token" debug messages.

IMO we should avoid fixing broken wikitext, even if we "fix" them the parsed nodes may still be unusable.

kristian-clausal:
Yeah, 'fixing' these so that they're like they "should" be based on some heuristics is probably not good. BUT we can make them *fail* in the same way they fail in Wiktionary; this, plus the fact that the .largs is missing the header text is what's the problem here, the heading doesn't have a heading at all.

Trying out on the Wikipedia sandbox, we've got:

```
=control 1=   // level 1

==control 2==  // level 2

=test==  // level 1, "test="

==test2===  // level 2, "test=="

=test  // normal text "=test"

[...]

PR #353, opened 2024-12-18
Heading tokens `==` should be parsed from outside in.

kristian-clausal:
Wikimedia parses heading tokens from inside out:
== Foo = is parsed as  =, "= Foo ", =, with leftover `=` characters in-between the bookends ending up in the the heading itself. Fixes issue #352.

Fixed by in the portion of the code where we tokenize headings. Happily, this is already done with a regex matching a whole line, so modifying the contents of the match is easy.

I will now fix and make tests. Fails one test because the output is different from expected (mismatched `=` now go inside the heading instead of being children of the level).

### DIFF
--- a/src/wikitextprocessor/parser.py
+++ b/src/wikitextprocessor/parser.py
@@ -2213,10 +2213,22 @@ def token_iter(ctx: "Wtp", text: str) -> Iterator[tuple[bool, str]]:
                 # the heading itself. Fixes issue #352.
                 start, mid, _, end = hm.groups()
                 if len(start) < len(end):
-                    mid += end[len(start):]
+                    ctx.debug(
+                        f"Heading `{start}`, `{mid}`, `{end}` "
+                        f"has a start token shorter than end token: "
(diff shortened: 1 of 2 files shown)


xxyzz:
Brilliant! I was only think the code in `subtitle_end_fn()`, this is much better.

kristian-clausal:
Yeah, luckily sometimes you can find a simple place to do something. We were already doing a lot of stuff at this point with special heading tokenizing, so it was a natural fit.

Tatu also suggested that it would be possible to modify header_re

```python
header_re = re.compile(r"(?m)^(={1,6})\s*(([^=]|=[^=])+?)\s*(={1,6})\s*$")
```

so that it would be a longer, dumber regex with `|`-separate options that would check `=(=====(....)(======))|(====(...)(=====))|` which would do the same thing, but we both suspect the performance difference between that and this wouldn't be very big. Thinking about it, I'm not sure how re would handle the capture groups in this case.

PR #356, opened 2025-01-06
Update Thai Wiktionary namespace file

xxyzz:
created with tools/get_namespaces.py and some manual translations


xxyzz:
Eh, the new mediawiki_langcodes release has some more data...

PR #361, opened 2025-01-28
Do not split pagenames with slashes in `Main:`

kristian-clausal:
Wiktextract issue https://github.com/tatuylonen/wiktextract/issues/1009

Pagename data passed into templates or modules was split on slashes, so you'd get output like `Bing`
and `Bed` instead of `A/Bing` and `A/Bed`.

We had some copied code from Scribunto handling
slashes in titles, but it didn't do a check for
whether slashes should be considered 'subpages',
like Module:mainmodule/submodule. `Main:` which
is our main concern, doesn't have subpages and
slashes should be considered part of the title.

I've only done a quick and dirty fix here, because we only concern ourselves with Main:, Template:
and Module:; All other should hopefully have
subpages (default assumption except for Main:)
or not have slashes in their titles.

### DIFF
--- a/src/wikitextprocessor/data/en/namespaces.json
+++ b/src/wikitextprocessor/data/en/namespaces.json
@@ -2,6 +2,7 @@
   "Media": {
     "id": -2,
     "name": "Media",
+    "subpages": false,
     "content": false,
     "aliases": [],
     "issubject": true,
@@ -10,6 +11,7 @@
   "Special": {
     "id": -1,
     "name": "Special",
+    "subpages": false,
     "content": false,
     "aliases": [],
     "issubject": true,
@@ -18,6 +20,7 @@
   "Main": {
     "id": 0,
     "name": "Main",
+    "subpages": false,
     "content": true,
     "aliases": [],
     "issubject": true,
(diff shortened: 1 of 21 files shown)


kristian-clausal:
Why don't I check make test...

kristian-clausal:
Apparently Main: pages can have subpages, like [head/translations](https://en.wiktionary.org/wiki/head/translations)... So this is not correct. OR possibly that the tests are wrong and the output shouldn't striped `/translations` from the 'title'.

xxyzz:
I only updated two tests, I also test the "A/B" page and the extracted forms are correct.

src/wikitextprocessor/data/en/namespaces.json
```diff
@@ -18,6 +20,7 @@
   "Main": {
     "id": 0,
     "name": "Main",
+    "subpages": false,
```
xxyzz, line 23: Might be little confusing, I guess they consider "Template:a/doc" a subpage but not "eye/translations".

kristian-clausal:
I think we don't access translations pages by partitioning "xxx/translations", instead of do "xxx" + "/translations", so this shouldn't, hopefully break anything. Thank you for checking this!

xxyzz:
I've tested the "eye" page in en edition and the translation data in "eye/translations" page are extracted correctly. Our Python code shouldn't be affected, just hope all that Lua code also works.

PR #363, opened 2025-02-13
Greek wiktionary uses '0', '00' and '000' as parameters

kristian-clausal:
https://el.wiktionary.org/w/index.php?title=%CE%A0%CF%81%CF%8C%CF%84%CF%85%CF%80%CE%BF:%CE%B5%CF%84

```
-->|0={{{0|}}} <!-- no parenthesis 
-->|00={{{00|}}} {{#if:{{{nocat|}}}|{{{nocat|}}}}}<!--
-->|000={{{000|}}} {{#if:{{{nodisplay|}}}|{{{nodisplay|}}}}}<!--
```

Apparently parameter names under 1, like '0' or '00' aren't changed to integers... Or that Scribunto handles all parameter names as strings and just does something special with the indexing stuff.

Still need to figure out some things...

### DIFF
--- a/src/wikitextprocessor/core.py
+++ b/src/wikitextprocessor/core.py
@@ -1212,7 +1212,7 @@ class Wtp:
                             expand_args(args[0], argmap), parent, True
                         ).strip()
                         self.expand_stack.pop()
-                        if k.isdigit():
+                        if k.isdigit() and int(k) > 0:
                             k = int(k)
                         else:
                             k = re.sub(r"\s+", " ", k).strip()
@@ -1428,7 +1428,7 @@ class Wtp:
                             # https://en.wikipedia.org/wiki/Help:Template
                             # (but not around unnamed parameters)
                             k, arg = m2.groups()
-                            if k.isdigit():
+                            if k.isdigit() and int(k) > 0:
                                 k = int(k)
                             else:
                                 self.expand_stack.append("ARGNAME")
--- a/src/wikitextprocessor/lua/_sandbox_phase2.lua
+++ b/src/wikitextprocessor/lua/_sandbox_phase2.lua
@@ -10,22 +10,37 @@ assert(_new_loadData ~= nil)
 
 local function frame_args_index(new_args, key)
     -- print("frame_args_index", key)
-    local i = tonumber(key)
-    if key ~= "inf" and key ~= "nan" and i ~= nil then
-        key = i
-    end
     local v = new_args._orig[key]
     if v == nil then
-        return nil
+        local i = tonumber(key)
+        if i ~= nil then
+            key = i
+        else
+            return nil
+        end
+        v = new_args._orig[key]
+        if v == nil then
+            return nil
+        end
     end
     if not new_args._preprocessed[key] then
         local frame = new_args._frame
-        v = frame:preprocess(v)
+        if type(v) == "userdata" then
+            -- Python tuple in luaexec.call_lua_sandbox.make_frame()
+            local is_named = v[1]
+            v = frame:preprocess(v[0])
+            -- https://en.wikipedia.org/wiki/Help:Template#Whitespace_handling
+            if is_named then
+                v = v:match "^%s*(.-)%s*$"
+            end
+        else
+            v = frame:preprocess(v)
+        end
         -- Cache preprocessed value so we only preprocess each argument once
         new_args._preprocessed[key] = true
         new_args._orig[key] = v
     end
-    -- print("frame_args_index", key, "->", v)
+    -- print("frame_args_index", key, "->", "'"..v.."'")
     return v
 end
 
--- a/src/wikitextprocessor/lua/mw.lua
+++ b/src/wikitextprocessor/lua/mw.lua
@@ -83,7 +83,90 @@ function mw.clone(v)
 end
 
 function mw.dumpObject(obj)
-    print("mw.dumpObject", obj)
+    -- https://github.com/wikimedia/mediawiki-extensions-Scribunto/blob/184216759e22635bd25a0844e6f68979ecf7bc2a/includes/Engines/LuaCommon/lualib/mw.lua#L551
+    local doneTable = {}
+    local doneObj = {}
+    local ct = {}
+    local function sorter( a, b )
+        local ta, tb = type( a ), type( b )
+        if ta ~= tb then
+            return ta < tb
+        end
+        if ta == 'string' or ta == 'number' then
+            return a < b
+        end
+        if ta == 'boolean' then
+            return tostring( a ) < tostring( b )
+        end
+        return false -- Incomparable
+    end
+    local function _dumpObject( object, indent, expandTable )
+        local tp = type( object )
+        if tp == 'number' or tp == 'nil' or tp == 'boolean' then
+            return tostring( object )
+        elseif tp == 'string' then
+            return string.format( "%q", object )
+        elseif tp == 'table' then
+            if not doneObj[object] then
+                local s = tostring( object )
+                if string.sub(s, 1, 6) == 'table:' then  -- this line changed
+                    ct[tp] = ( ct[tp] or 0 ) + 1
+                    doneObj[object] = 'table#' .. ct[tp]
+                else
+                    doneObj[object] = s
+                    doneTable[object] = true
+                end
+            end
+            if doneTable[object] or not expandTable then
+                return doneObj[object]
+            end
+            doneTable[object] = true
+
+            local ret = { doneObj[object], ' {\n' }
+            local mt = getmetatable( object )
+            local indentString = "  "
+            if mt then
+                ret[#ret + 1] = string.rep( indentString, indent + 2 )
+                ret[#ret + 1] = 'metatable = '
+                ret[#ret + 1] = _dumpObject( mt, indent + 2, false )
+                ret[#ret + 1] = "\n"
+            end
+
+            local doneKeys = {}
+            for key, value in ipairs( object ) do
+                doneKeys[key] = true
+                ret[#ret + 1] = string.rep( indentString, indent + 2 )
+                ret[#ret + 1] = _dumpObject( value, indent + 2, true )
+                ret[#ret + 1] = ',\n'
+            end
+            local keys = {}
+            for key in pairs( object ) do
+                if not doneKeys[key] then
+                    keys[#keys + 1] = key
+                end
+            end
+            table.sort( keys, sorter )
+            for i = 1, #keys do
+                local key = keys[i]
+                ret[#ret + 1] = string.rep( indentString, indent + 2 )
+                ret[#ret + 1] = '['
+                ret[#ret + 1] = _dumpObject( key, indent + 3, false )
+                ret[#ret + 1] = '] = '
+                ret[#ret + 1] = _dumpObject( object[key], indent + 2, true )
+                ret[#ret + 1] = ",\n"
+            end
+            ret[#ret + 1] = string.rep( indentString, indent )
+            ret[#ret + 1] = '}'
+            return table.concat( ret )
+        else
+            if not doneObj[object] then
+                ct[tp] = ( ct[tp] or 0 ) + 1
+                doneObj[object] = tostring( object ) .. '#' .. ct[tp]
+            end
+            return doneObj[object]
+        end
+    end
+    return _dumpObject( obj, 0, true )
 end
 
 function mw.incrementExpensiveFunctionCount()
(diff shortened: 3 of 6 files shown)


kristian-clausal:
I am trying to fix a Lua error, and I think I've figured out what's wrong:

```
$ lang=el; datadir=/home/kristian/Data/${lang};wiktwords --edition "$lang" --db-path "${datadir}/${lang}-wikt.db" --all --human-readable --out temp --errors temp.errors --all-languages --page gtmp/thelukos2.txt --override el-override/
2025-02-13 11:31:59,481 INFO: Capturing words for all available languages
gl: "           " lang: "       nil
*32*    table: 0x341b73c0
θηλυκός: ERROR: LUA error in #invoke('labels', 'main', '', '', 'label=γραμμ \U0010207b \U0010207b \U0010207b ', 'όρος= \U0010207b ', 'εμφ= \U0010207b ', 'σελ= \U0010207b  ', '0= ', '00=- \U0010207b', '000= \U0010207b', 'γλ= \U0010207b \U0010207b', 'ascii= \U0010207b \U0010207b') parent ('Πρότυπο:ετ', {1: 'γραμμ', '00': '-'}) at ['θηλυκός', 'Template:ετ', '#invoke', '#invoke']
[string "Module:labels"]:71: attempt to index field '?' (a nil value)
gl: "           " lang: "       nil
*32*    table: 0x341b73c0
θηλυκός: ERROR: LUA error in #invoke('labels', 'main', '', '', 'label=γραμμ \U0010207b \U0010207b \U0010207b ', 'όρος= \U0010207b ', 'εμφ= \U0010207b ', 'σελ= \U0010207b  ', '0= ', '00=- \U0010207b', '000= \U0010207b', 'γλ= \U0010207b \U0010207b', 'ascii= \U0010207b \U0010207b') parent ('Πρότυπο:ετ', {1: 'γραμμ', '00': '-'}) at ['θηλυκός', 'Template:ετ', '#invoke', '#invoke']
[string "Module:labels"]:71: attempt to index field '?' (a nil value)
```
In Greek wiktionary Module:labels we have

```lua
-- about languages, language specifics CHECK [[takeout]] [[φάλαινα]]
	print('gl: "', args['γλ'], '" lang: "', args['lang']) 
   	local lang_iso = args['γλ'] or args['lang'] or '' -- or args[2] at [[Template|ετ]] 
		if lang_iso == '' or lang_iso == nil then 
			if label == 'αμερ' or label == 'αμερ γρ' or label == 'αμερ σημασία'
			or label == 'βρετ' or label == 'βρετ γρ' or label == 'βρετ σημασία'
			then lang_iso = 'en'
			else lang_iso = 'el'
			end
		end
	print('*' .. string.byte(lang_iso) .. '*', languages)
	local lang_name = languages[lang_iso].name or ''
```
and it turns out that `args['γλ']` is one character of space, just ASCII 32. This doesn't seem to be related to the '0' stuff, though, just coincidental... No idea why letting spaces through doesn't happen more often, I'll try to investigate.

kristian-clausal:
```
[string "Module:labels"]:95: attempt to index field '?' (a nil value)
(venv) kristian@wiktextract-dev:~/Alt/wiktextract (el):
$ lang=el; datadir=/home/kristian/Data/${lang};wiktwords --edition "$lang" --db-path "${datad
2025-02-13 12:30:22,201 INFO: Capturing words for all available languages
NOT REACHED?    "γραμμ   "
gl: "           " lang: "       nil
*101*   table: 0xb500270
Label:  γραμμ
data[label]:    nil
θηλυκός: ERROR: LUA error in #invoke('labels', 'main', '', '', 'label=γραμμ \U0010207b \U0010207b \U0010207b', 'όρος= \U0010207b', 'εμφ= \U0010207b', 'σελ= \U0010207b', '0=', '00=- \U0010207b', '000= \U0010207b', 'γλ= \U0010207b \U0010207b', 'ascii= \U0010207b \U0010207b') parent ('Πρότυπο:ετ', {1: 'γραμμ', '00': '-'}) at ['θηλυκός', 'Template:ετ', '#invoke', '#invoke']
[string "Module:labels"]:95: attempt to index field '?' (a nil value)
NOT REACHED?    "γραμμ   "
gl: "           " lang: "       nil
*101*   table: 0xb500270
Label:  γραμμ
data[label]:    nil
θηλυκός: ERROR: LUA error in #invoke('labels', 'main', '', '', 'label=γραμμ \U0010207b \U0010207b \U0010207b', 'όρος= \U0010207b', 'εμφ= \U0010207b', 'σελ= \U0010207b', '0=', '00=- \U0010207b', '000= \U0010207b', 'γλ= \U0010207b \U0010207b', 'ascii= \U0010207b \U0010207b') parent ('Πρότυπο:ετ', {1: 'γραμμ', '00': '-'}) at ['θηλυκός', 'Template:ετ', '#invoke', '#invoke']
[string "Module:labels"]:95: attempt to index field '?' (a nil value)
```

aaaaargh I don't get why these template arguments don't get stripped properly. This is probably unrelated to the `000` issue: `γραμμ   ` with three spaces (because there are three spaces between three `{{{...|...}}}` spans in `|label={{{1|}}} {{#if:{{{label|}}}|{{{label|}}}}} {{#if:{{{topic|}}}|{{{topic|}}}}} {{#if:{{{ετικέτα|}}}` in `Επεξεργασία: Πρότυπο:ετ` means the Lua code can't find the correct data under `γραμμ` in `labels/data`... There is a step somewhere that does `.strip()` *somewhere*. Because `label=` is used in `Επεξεργασία: Πρότυπο:ετ` it should strip the stuff, so maybe it's a timing issue...

xxyzz:
We should be able to pass this test, but current we get  result of "nilnilnilnil":
```python
    def test_gl_zero_arg(self):
        self.wtp.start_page("Πρότυπο:ετ")
        self.wtp.add_page(
            "Module:test",
            828,
            """local export = {}
function export.test(frame)
  return tostring(frame.args[0]) .. tostring(frame.args["0"]) .. tostring(frame.args[00]) .. tostring(frame.args["00"])
end
return export""",
        )
        self.assertEqual(self.wtp.expand(
            "{{#invoke:test|test|0=0|00=1}}"), "0001"
        )
```

xxyzz:
I'd like to suggest these changes:

in "_sandbox_phase2.lua", try string key first then try number key:
```lua
    local v = new_args._orig[key]
    if v == nil then
        local i = tonumber(key)
        if key ~= "inf" and key ~= "nan" and i ~= nil then
            key = i
        end
        v = new_args._orig[key]
        if v == nil then
            return nil
        end
    end
```

in luaexec.py, only convert "0" and numbers that don't start with "0":
```python
if k.isdigit() and (not k.startswith("0") or k == "0"):
    k = int(k)
    if k < 0 or k > 1000:
```

xxyzz:
For removing white spaces in expanded template arg:

```python
    def test_el_strip_arg(self):
        self.wtp.start_page("θηλυκός")
        self.wtp.add_page(
            "Module:test",
            828,
            """local export = {}
function export.test(frame)
  return tostring(frame.args[0])
end
return export""",
        )
        self.assertEqual(self.wtp.expand(
            "{{#invoke:test|test|0={{{1|}}} {{{2|}}}}}"), ""
        )
```

at the end of `preprocess()` in luaexec.py:
```python
ret = expand_all_templates(v)
if ret.strip() == "":
    ret = ""
```

kristian-clausal:
I think that only positive, non-zero integers are allowed in Wikitext parameters, so using `int(k) > 0` should be correct.

The current problem is, I think, that when we do stuff in wikitextprocessor/core expand_recurse:

```python
                    if kind == "T":
                        # Template transclusion or parser function call.
                        # Expand its arguments.
                        print(f"{kind=}, {args=}, {argmap=}")
                        new_args = tuple(
                            expand_args(x, argmap).removesuffix("\n")
                            for x in args
                        )
                        print(f"{new_args=}")
                        parts.append(self._save_value(kind, new_args, nowiki))
                        continue
```

That seems to always leave a magic character for `T` expansions (including `#if`), and usually that magic character is the same value as all the other magic characters left because we keep on .pop()ing stuff from the magic character cookies stack and then putting something on top again and again. But there's not check for whether the template needs to have a cookie here.

```
new_args=('#invoke:labels', 'main', '', '', 'label=γραμμ \U0010207b \U0010207b \U0010207b ', 'όρος= \U0010207b ', 'εμφ= \U0010207b ', 'σελ= \U0010207b  ', '0= ', '00=- \U0010207b', '000= \U0010207b', 'γλ= \U0010207b \U0010207b', 'ascii= \U0010207b \U0010207b')
fn_name='#if', args=('', ''), ret=''
fn_name='#if', args=('', ''), ret=''
...
*101*   table: 0x1fa28c70
Label:  γραμμ
data[label]:    nil
θηλυκός: ERROR: LUA error in #invoke('labels', 'main', '', '', 'label=γραμμ \U0010207b \U0010207b \U0010207b ', 'όρος= \U0010207b ', 'εμφ= \U0010207b ', 'σελ= \U0010207b  ', '0= ', '00=- \U0010207b', '000= \U0010207b', 'γλ= \U0010207b \U0010207b', 'ascii= \U0010207b \U0010207b') parent ('Πρότυπο:ετ', {1: 'γραμμ', '00': '-'}) at ['θηλυκός', 'Template:ετ', '#invoke', '#invoke']
[string "Module:labels"]:95: attempt to index field '?' (a nil value)
```

What used to be normal magic characters with different values have been replaced by `\U0010207b`.


Why hasn't this come up before? I think it's because Lua simply discards the magic character because it's out of Unicode's normal range of characters. But in this case, all the `#if` calls in the template have spaces inbetween them and we're left with whitespace that isn't `.strip()`ped away as normal. All of this goes straight to the Lua interpreter, with the magic characters in place.

xxyzz:
I think the first "0" arg problem is not fixed, because `frame.args["0"]` is `nil` in this branch.

Could you add a simplified test for the second magic characters problem?

kristian-clausal:
Aaah, there's a check to see if the cookie data already exists in `_save_value` using `rev_ht`. Because the `#if` templates are all reduced to the same data `("#if", "", "")` or something like that, basically an empty if, _save_value returns the same. In https://el.wiktionary.org/w/index.php?title=%CE%A0%CF%81%CF%8C%CF%84%CF%85%CF%80%CE%BF:%CE%B5%CF%84 we have

```
-->|label={{{1|}}} {{#if:{{{label|}}}|{{{label|}}}}} {{#if:{{{topic|}}}|{{{topic|}}}}} {{#if:{{{ετικέτα|}}}|{{{ετικέτα|}}}}} <!-- forms Categories and/or link 
-->|όρος={{{όρος|}}} {{#if:{{{term|}}}|{{{term|}}}}} <!-- overrides the chosen text 
-->|εμφ={{{3|}}} {{#if:{{{show|{{{εμφ|}}}}}}|{{{show|{{{εμφ|}}}}}}}} <!-- shows chosen text at link εμφ π.χ. Μεγάλη Άρκτος
-->|σελ={{{σελ|}}} {{#if:{{{page|}}}|{{{page|}}}}}  <!-- σελ=1 συνδέει με το λήμμα, αντί την Κατηγορία - Links lemma, not default Category 
-->|0={{{0|}}} <!-- no parenthesis 
-->|00={{{00|}}} {{#if:{{{nocat|}}}|{{{nocat|}}}}}<!--
-->|000={{{000|}}} {{#if:{{{nodisplay|}}}|{{{nodisplay|}}}}}<!--
-->|γλ={{{2|}}} {{#if:{{{γλ|}}}|{{{γλ|}}}}} {{#if:{{{lang|}}}|{{{lang|}}}}}<!--
-->|ascii={{{ascii|}}} {{#if:{{{ascii|}}}|{{{ascii|}}}}} {{#if:{{{sort|}}}|{{{sort|}}}}}<!--
```

and it's exactly those `#if`s that seem to remain as cookies.

kristian-clausal:
The magic characters end up getting into Lua. They're not expanded. Lua just seems to discard the weird Unicode characters. At least it seems that way...

kristian-clausal:
`sandbox_phase2.lua` starting line 21
```lua
    if not new_args._preprocessed[key] then
        local frame = new_args._frame
        v = frame:preprocess(v)
        if key ~= i then
            v = v:match "^%s*(.-)%s*$"
        end
        -- Cache preprocessed value so we only preprocess each argument once
        new_args._preprocessed[key] = true
        new_args._orig[key] = v
```

This seems to do it. The check for `key ~= i` is because we want to know if this was an argument that used `aa=bb`, because those don't have their spaces trimmed (according to documentation); if it's just `|    bb   |` then that should be trimmed. I'm not sure if `1=   bb   ` would be trimmed...

xxyzz:
"1=   bb " do get striped... and don't forget `frame.args["0"]` shouldn't be `nil`.

kristian-clausal:
@xxyzz I had to make changes to one of your tests (Russian, remove \n at the end of unnamed arguments) and I would like you to check I didn't do anything dumb again.

kristian-clausal:
Thanks for taking a look and completing the stuff I missed! :+1:

PR #374, opened 2025-04-16
Tests to fail `template_parameters`

kristian-clausal:
`TemplateNode.template_parameters` has a broken return value.

If you go to the [enwiktionary sandbox]() and enter

```
{{/Please do not edit this line}}
{{emojipic|u1f384|🎄}}
{{emojipic|🎄|1=u1f384}}
{{emojipic|2=🎄|u1f384}}
```

We can see that it works correctly, but currently the `.template_parameters` implementation is wrong in this case.

We are doing maintenance here at work today and the power is cut soon so I'm committing these tests first.

### DIFF
--- a/tests/test_wikinode.py
+++ b/tests/test_wikinode.py
@@ -1,0 +1,35 @@
+# Tests for WikiText parsing
+#
+# Copyright (c) 2020-2021 Tatu Ylonen.  See file LICENSE and https://ylonen.org
+
+import unittest
+
+from wikitextprocessor import Wtp
+
+
+class WikiNodeTests(unittest.TestCase):
+    def setUp(self):
+        self.ctx = Wtp()
+
+    def tearDown(self):
+        self.ctx.close_db_conn()
+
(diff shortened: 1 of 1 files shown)


xxyzz:
`{{emojipic|🎄|1=u1f384}}` is expanded to `[[File:Emoji u1f384.svg|thumb||200px|The [[emoji]] for a: see the entry at [[{{{2}}}]].]]` on Wiktionary, notice the unexpanded `{{{2}}}`, so it's not supported by MediaWiki. I doubt anyone would write arguments in this way intentionally...

Other tests are passed and already have similar tests in `test_parser.py` with real usage like https://github.com/tatuylonen/wikitextprocessor/blob/ea19a0a7c5b481aa68246c6ccff1c86eb8f8be26/tests/test_parser.py#L2609-L2611

kristian-clausal:
Why did you close this? This is obviously incorrect in the code, and should be fixed!

PR #381, opened 2025-05-15
Don't close list node if new line is inside HTML tag

xxyzz:
en page "bump" in issue tatuylonen/wiktextract#1196

### DIFF
--- a/src/wikitextprocessor/parser.py
+++ b/src/wikitextprocessor/parser.py
@@ -862,6 +862,11 @@ def close_begline_lists(ctx: "Wtp") -> None:
     """Closes currently open list if at the beginning of a line."""
     if not (ctx.beginning_of_line and ctx.begline_enabled):
         return
+    # only check if the last direct parent node is HTML
+    # if list doesn't close properly, update this code
+    node = ctx.parser_stack[-1]
+    if isinstance(node, HTMLNode) and node.tag in ctx.paired_html_tags:
+        return
     while _parser_have(ctx, NodeKind.LIST):
         _parser_pop(ctx, True)
 
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -3207,6 +3207,32 @@ text
         self.assertEqual(level_2.kind, NodeKind.LEVEL2)
         self.assertEqual(level_2.children[1].kind, NodeKind.LEVEL4)
 
+    def test_new_line_in_ref(self):
+        # shouldn't break ref tag and list
+        root = self.parse(
+            "bump",
+            """<div>
+# To <ref>
+text</ref>
+</div>""",
+        )
+        div_node = root.children[0]
+        self.assertEqual(len(div_node.children), 2)
+        self.assertEqual(div_node.children[0], "\n")
+        self.assertEqual(div_node.children[1].kind, NodeKind.LIST)
+        # list and span should be sibling
+        root = self.parse(
+            "bump",
+            """<div>
+# List
+<span>text</span>
+</div>""",
+        )
+        div_node = root.children[0]
+        self.assertEqual(len(div_node.children), 4)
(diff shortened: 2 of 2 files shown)


src/wikitextprocessor/parser.py
```diff
@@ -862,6 +862,9 @@ def close_begline_lists(ctx: "Wtp") -> None:
     """Closes currently open list if at the beginning of a line."""
     if not (ctx.beginning_of_line and ctx.begline_enabled):
         return
+    node = ctx.parser_stack[-1]
+    if isinstance(node, HTMLNode) and node.tag in ctx.paired_html_tags:
```
kristian-clausal, line 866: Could it be possible that the `HTMLNode` is found in `ctx.parser_stack[-2]` or deeper in the stack? If we have

```
# To <ref>'''bold
text'''</ref>
```

or something similar. We might need to check if there's an open HTML inside a list.

xxyzz: The updated code could handle this case but I didn't update the test because bold wikitext shouldn't have new line in it...

src/wikitextprocessor/parser.py
```diff
@@ -862,6 +862,9 @@ def close_begline_lists(ctx: "Wtp") -> None:
     """Closes currently open list if at the beginning of a line."""
     if not (ctx.beginning_of_line and ctx.begline_enabled):
         return
+    for node in reversed(ctx.parser_stack):
```
kristian-clausal, line 865: This doesn't take into account if the LIST is inside a larger HTML node.

```
<div>
# List
</div>
```

xxyzz:
```
<div>
# List<ref>
text</ref>
</div>
```

this is parsed correctly as `<ROOT(['bump']){} <HTML(div){} '\n', <LIST(#){} <LIST_ITEM(#){} ' List', <HTML(ref){} '\ntext'>, '\n'>>>>`

kristian-clausal: I meant that if the LIST is inside a larger HTML node, it breaks because it will eventually find the outside node and return when it shouldn't.

```
<div>
# List
</div>
```

The list will not be closed here. I think, I hope I didn't get the logic wrong.

xxyzz: Could you come up a test the code parse incorrectly? I also tested
```
<div>
# List
text
</div>
``` 
and it seems fine, maybe it's closed at other place?

kristian-clausal: You're right, I can't make it break! I hate no understand why it doesn't break...

kristian-clausal (approved):

xxyzz:
Don't merge now...

xxyzz:
I find the commit added this function: https://github.com/tatuylonen/wikitextprocessor/commit/2519cf65749d31d9eae9f5d71858daec1def8e6c

I think it was created for this kind of nodes:
```
<div>
# List
<span>text</span>
</div>
```

the list and span should be parsed as two sibling nodes. But the current code doesn't close the list and span is child of the list. If I revert to the first version only check the last node in stack, it'll be parsed correctly.

kristian-clausal:
Oh, that was it! Either we revert to the simple "check the last node" version, or we have to check if we're inside an HTML node that is inside a LIST node by checking the span of the stack. The first might break if the HTML contains some other nodes (unlikely?), but the other is so much slower...

xxyzz:
IDK... only check the last node fix the current ref new line bug and also seems to fix the original bug. Hopefully it won't break some code.

kristian-clausal:
Add a comment there that says something like "If this causes more problems in the future, do this more complicated check", so that future-us will know we were doing it on purpose, and that should do it. It seems unlikely it will cause problems.

PR #384, opened 2025-07-28
Fix `ZeroDivisionError` exception in `padleft_fn`

xxyzz:


kristian-clausal:
You're quick, I'm still waiting for the dump to be extracted into the .db file :D

kristian-clausal (approved):

xxyzz:
Recent years PC CPU's single core speed is twice of the server CPU...

PR #385, opened 2025-07-28
Also apply changes in parser function `padleft` to `padright`

kristian-clausal:


kristian-clausal:
I just copied the changes and tests (and the behavior in Wiktionary:sandbox seems to be the same as in the tests), if you could quickly see if I messed something up because of the direct copying.

xxyzz (approved):

PR #386, opened 2025-07-31
Downgrade dateparser 1.2.1

xxyzz:
1.2.2 has a broken pickle cache fails "test_time34".


kristian-clausal:
Thanks for taking care of this!

PR #388, opened 2025-08-21
fix: commit db changes on close

BoboTiG:
Without this change, the DB was always empty at the end of my tests for [reader.dict](https://github.com/reader-dict).


xxyzz:
You're extracting Wiktionary? Why not using our data https://kaikki.org/dictionary/rawdata.html ?

kristian-clausal:
Is it possible for this to cause future side-effects in tests?

xxyzz:
> Is it possible for this to cause future side-effects in tests?

I think it shouldn't affect tests.

PR #394, opened 2025-09-18
Allow pipe character in external links

xxyzz:
fix Lua error in cs edition, "|" in external links should be parsed as plain text


kristian-clausal (approved):
Oh, wow. I'm not sure why I'm surprised we still keep on finding edge cases like this. Thanks for taking care of this!

PR #398, opened 2025-10-16
Add new error message categories: NOTE and WIKI

kristian-clausal:
NOTE is for general messages that don't fit in debug, and isn't really a warning or error.

WIKI is for everything that's wrong on the Wiki source side of things, in cases where we have already corrected the issue on our side. Wiktionary errors in arguments, etc.


kristian-clausal:
I am using ruff as a formatter, but the output is different. Ugh. Difference in versions, updating ruff in pip seems to give the same output.

PR #399, opened 2025-10-21
`||` inside table should be parsed as one cell

xxyzz:
On Wiktionary, `||` at the line beginning inside a table is expanded to a single `<td>` HTML element. Previous code produces an extra empty table cell node and makes it more difficult for extractor code to decide whether to discard the cell.

`test_table_hdr4` was added in commit 1f48e90 for en edition page "山 歩き" and the extracted forms list is the same after this change.


kristian-clausal:
You might need to upgrade ruff, I had the same problem with the linter being mean to me. The format specs have been changed.

xxyzz:
I forget to run `ruff format`...

PR #403, opened 2025-11-04
Don't parse HTML attribute as switch parser function argument

xxyzz:
found the bug in ru edition page "больной"

### DIFF
--- a/src/wikitextprocessor/parserfns.py
+++ b/src/wikitextprocessor/parserfns.py
@@ -111,10 +111,10 @@ def switch_fn(
     next_val_is_default = False
     defval: Optional[str] = None
     last: Optional[str] = None
-    for i in range(1, len(args)):
-        arg = args[i]
+    for arg in args[1:]:
         m = re.match(r"(?s)^([^=]*)=(.*)$", arg)
-        if not m:
+        # "=" in HTML attribute is not switch argument
+        if m is None or ("<" in m.group(1) and ">" in m.group(2)):
             last = expander(arg).strip()
             if last == val:
                 match_next = True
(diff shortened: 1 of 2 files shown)


src/wikitextprocessor/parserfns.py
```diff
@@ -114,7 +114,8 @@ def switch_fn(
     for i in range(1, len(args)):
```
kristian-clausal: You can tell this is old code, it's before Tatu found about `enumerate`.

src/wikitextprocessor/parserfns.py
```diff
@@ -114,7 +114,8 @@ def switch_fn(
...
         arg = args[i]
         m = re.match(r"(?s)^([^=]*)=(.*)$", arg)
-        if not m:
+        # "=" in HTML attribute is not switch argument
+        if m is None or ("<" in m.group(1) and ">" in m.group(2)):
```
kristian-clausal: AFAICT, this kind of confusion can only happen in the last, un-named argument of a `#switch`:

`{{#switch: {{{x}}} |1=one |2=two |3|4|5=range 3–5 |other}}`, in the `other`.

We could do `... or (if i == len(args) -1 and "<" in m.group(1) ...)`, I think that should be correct? But yeah, I can't figure out a good way to tell if it's an HTML element that isn't just parsing the last argument.

kristian-clausal (approved):

xxyzz:
Perhaps we could reuse this template arg regex? https://github.com/tatuylonen/wikitextprocessor/blob/f683e8a1726a549dd3114e50663214c316753887/src/wikitextprocessor/core.py#L1485

PR #404, opened 2025-11-04
Minor improvement of previous pr #403

xxyzz:
copy part of regex used for template args in "core.py" line 1485


kristian-clausal (approved):
LGTM!

PR #405, opened 2025-11-05
Fix `index out of range` exception in en page "aderedor"

xxyzz:
only check direct HTML parent node is not enough


xxyzz:
I also think it's safer to only check ref tag.

PR #416, opened 2026-04-01
Add option to not print output

Darkdadaah:
Add 'quiet_output' to Wtp to not print errors to stdout via _fmt_errmsg()

When parsing a dump the output from errors (`Wtp.error()` and the likes) can be quite overwhelming, and not needed when all the errors are accessible via `Wtp.to_return()` in a parsable data structure. This new option to `Wtp` helps keep the output quiet when it is not needed.
I considered using the `quiet` parameter, but its current use is to quiet the logging. Better not mix things up I suppose.

Usage:
```
wtp = Wtp(quiet_output=True)
```
(Defaults to False, current behavior)

NB: the error messages sent to stdout cause a problem with unittest for `wiktextractor` where the parser is run in threads, in which case the stdout of the threads can't be captured by unittest. This PR is an attempt to mitigate that problem without drastic changes to the current behavior of the code.

### DIFF
--- a/src/wikitextprocessor/core.py
+++ b/src/wikitextprocessor/core.py
@@ -283,6 +283,7 @@ class Wtp:
         "wiki_notices",  # WIKI error messages
         "wikidata_session",
         "linktrailing_re",
+        "quiet_output", # Prevent errors printed to stdout
     )
 
     def __init__(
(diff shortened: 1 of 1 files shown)


kristian-clausal:
This seems fine, only the linter complains.

PR #429, opened 2026-09-13
feat: update MG namespaces

BoboTiG:
Without this sync, impossible to render a single template.

Updated using `tools/get_namespaces.py`.


kristian-clausal:
Thank you for this update!

PR #432, opened 2026-09-16
[feat] Implement PHP doQuotes logic

daxida:
Convoluted because their logic is (and tbh, rightfully so). The myriad of tests and references to the original PHP code should be good enough (they don't even cover all of the edge cases of their implementation, so I didn't try it either!)

See https://github.com/tatuylonen/wiktextract/issues/1706

The test I added in https://github.com/tatuylonen/wiktextract/pull/1707 should pass when this lands, and we will have validation then to close that issue.

For full transparency, the code was LLM generated and only marginally tweaked by hand. I did check the implementation a couple times and really nudged it to simplify much of the initial boilerplate/noisy LLM output we all know and love.


src/wikitextprocessor/parser.py
```diff
@@ -2208,15 +2208,66 @@ def magicword_fn(ctx: "Wtp", token: str) -> None:
...
+        if parts[i] != "'''":
             continue
-        if p.startswith("'''"):
-            return True
-    return False
+        before = parts[i - 1]
+        if before.endswith(" "):
+            if first_space < 0:
+                first_space = i
+        elif before[-2:-1] == " ":
```
kristian-clausal, line 2257: This seems to check for "foo l", but what about if the single letter is the whole string? "L'foobar".

src/wikitextprocessor/parser.py
```diff
@@ -2281,83 +2332,33 @@ def token_iter(ctx: "Wtp", text: str) -> Iterator[tuple[bool, str]]:
                 yield True, ">" + start
             continue
         # Partition on ''+, so that we can detect bold/italics
-        parts = re.split(parts_re, line)
+        parts = balance_apostrophes(re.split(parts_re, line))
```
kristian-clausal, line 2335: Nothing to do with this review, I'm just reminded that we should really take all of the regex patterns and make them proper consts outside of the function they're defined in. There's a few places where we do actually manipulate regex patterns on the fly, but this one I should clean up next time I touch this part of the code.

src/wikitextprocessor/parser.py
```diff
@@ -2281,83 +2332,33 @@ def token_iter(ctx: "Wtp", text: str) -> Iterator[tuple[bool, str]]:
...
-                if part:
-                    # Shouldn't contain MAGIC_SQUOTE_CHAR
-                    yield False, part
+            # Markers are at the odd indices; go by index, since an even part
+            # can now start with apostrophes that are literal text.
+            if i % 2 == 1:
+                # Each marker is 2, 3 or 5 apostrophes: italic, bold, both.
+                if len(part) != 5:
+                    yield True, part
+                    state ^= 1 if len(part) == 2 else 2
```
kristian-clausal, line 2347: Please leave a comment explaining what's happening with the bitwise stuff here. Should probably be written in a more idiomatic way.

src/wikitextprocessor/parser.py
```diff
@@ -2281,83 +2332,33 @@ def token_iter(ctx: "Wtp", text: str) -> Iterator[tuple[bool, str]]:
...
+                    continue
+                # Inner tag first: italic if it closes, or if bold outlives it.
+                if state == 1 or (state == 0 and bold_follows(parts, i)):
+                    yield True, "''"
+                    yield True, "'''"
+                else:
+                    yield True, "'''"
+                    yield True, "''"
+                state ^= 3
```
kristian-clausal, line 2356: Same with the xor assignment here.

kristian-clausal (changes requested):

src/wikitextprocessor/parser.py
daxida: This and the other XOR follow the same logic.

```py
state = 0  # 1=in italic, 2=in bold, 3=in both
```
or, in other words, two real states A (italics) and B (bold), then 0 for (not A and not B) and 3 for (A and B).

The states lend themselves nicely to bitwise operations: XOR with 1 means "flip the italics bit", XOR with 2 means "flip the bold bit", XOR with 3 means "flip both".

Something like 

```py
state ^= 1 if len(part) == 2 else 2
```
means. if we have two parts, flip italics, else flip bold etc.

I guess I could add a comment, but I could also make the state into an enum with methods like "flip_italics" and such, I don't think it would have much overhead and it should be probably clearer, thoughts?

---

A table for visualization

| ⊕ | 0 | 1 | 2 | 3 |
|:-:|:-:|:-:|:-:|:-:|
| **0** | 0 | 1 | 2 | 3 |
| **1** | 1 | 0 | 3 | 2 |
| **2** | 2 | 3 | 0 | 1 |
| **3** | 3 | 2 | 1 | 0 |

src/wikitextprocessor/parser.py
daxida: I'm not sure I see your point, do you want me to add a test for this? Do you have something concrete in mind (input/expected output?). 

Note that I ended up removing a bunch of other edge case tests made by the LLM because they weren't even tested upstream anyway...

kristian-clausal: I mean that this is trying to find the first single-letter word, but the condition using a space will miss a single letter word at the start of a string. It will find `foo l'bar`, but miss `L'bar`

src/wikitextprocessor/parser.py
kristian-clausal: Please add a quick comment about what the code is doing in the sense of "flip bold bit". We're only doing this in this relatively small part of the code, so enums don't feel completely necessary and I'd rather avoid even just a little bit of overhead; I'm sure the cost is just seconds or minutes of run-time for a whole wiktextract run, but this thing is such a lumbering beast already...

src/wikitextprocessor/parser.py
daxida: Ah yeah, the tldr is PHP does this too. I guess we could upstream it but yeah...

L'''a'' b x'''c'''d

Captura de pantalla 2026-09-17 a las 8 08 28
Captura de pantalla 2026-09-17 a las 8 09 16

src/wikitextprocessor/parser.py
daxida: Well I'll add a comment in the state initialization before the loop, if that is fine with you (instead of at each XOR),

src/wikitextprocessor/parser.py
kristian-clausal: Oh! If upstream does this, then we can just leave it as is here, and maybe wait for things to change upstream. As long as we are as consistent as possible with what the wikitext output is, it's fine; in these kinds of situation, *we* are not able to catch weird stuff like this when using wikitextprocessor easily, but from the point of view of wikitext using normal users this is a quirk that is visible in the output and that they might even fix, which means the errors are less likely to "propagate" down to us because the source is more likely to be correct.

daxida:
Let me know if the wording of the comment suits you o/

kristian-clausal (approved):
Looks good to me.

kristian-clausal:
Thank you for the commit!

